PR #22, opened 2019-04-23
Seqyclean Dockerfile

Ikkik:
I think it works!


kapsakcj:
Yup, built with no issues and I ran the test PE reads through and worked great. Thanks for making this docker image!

I'll make a docker hub repo, add you to the contributors list, and update the README

PR #33, opened 2019-07-30
added downloading the serotypefinder database

Ikkik:
Copied original dockerfile and added downloading the serotypefinder database

TL;DR 
We're trying to use serotypefinder to determine E. coli subtype, as opposed to the EcOH database, but we like using abricate, so I thought it'd be awesome to add serotypefinder to this container of abricate.


k-florek:
Merging update with master

PR #41, opened 2019-08-20
Adding directory for kraken databases

Ikkik:


kapsakcj:
Didn't realize that Mash used perl! I'll push these changes to dockerhub shortly.

PR #47, opened 2019-10-30
set perl locale settings...

Ikkik:
Canu-racon container is giving me the following error when run as a singularity container:

```
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
```


kapsakcj:
Thanks for catching this! Guess you're the first to use the canu-racon container using singularity.

I'll rebuild on dockerhub shortly

kapsakcj:
@Ikkik build has finished on dockerhub. Try it out and let me know how it goes

PR #50, opened 2019-11-12
Added pandoc for pdf output

AbigailShockey:
Tested and multiqc can now output pdfs


kapsakcj:
gonna rebuild on dockerhub now

kapsakcj:
Docker image should be ready in ~20 min or so

PR #86, opened 2020-03-18
Add Snippy tag with SC2 reference genome for reference-based assembly

kevinlibuit:


snippy/4.6.0_SC2/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+# Samclip 0.4 - bundled
+
+# snp-sites 2.3.3 - via apt-get 
+
+# any2fasta 0.4.2 - bundled
+
+# wgsim 0.3.1-r13 - bundled for testing snippy
+
+# SARS-CoV-2 Reference Genome
+RUN curl -s ftp://ftp.ncbi.nlm.nih.gov/genomes/all/GCF/009/858/895/GCF_009858895.2_ASM985889v3/GCF_009858895.2_ASM985889v3_genomic.fna.gz | gzip -dc > MN908947_SC2_reference.fasta
```
kapsakcj: Are these two lines needed? Looks like you download the same thing below, with the difference being that its put into `/reference`

kevinlibuit: Ah, sorry about that. Sloppy PR. These lines aren't needed. I wrote it like this the first time then decided to consolidate things into a single RUN statement.

kapsakcj: no worries, just checking to make sure it wasn't for a specific reason! Can you delete the first two in a new commit? I'll merge and push to dockerhub afterwards

kapsakcj:
Any thoughts on what you want the dockerhub tag to be? `staphb/snippy:4.6.0-SC2` ?

I don't remember offhand what the syntax requirements are for dockerhub tags, but I'd like to keep the snippy version in there and keep it concise

kapsakcj:
Also -- ignore travis CI failing. It's failing for an unrelated reason. I'll merge the PR regardless

kevinlibuit:
I like staphb/snippy:4.6.0-SC2 tag. If this is the naming convention in dockerhub, we should change the subdir in the docker-biulds repo as well; I wrote it as "snippy/4.6.0_SC2"

kapsakcj:
Dockerhub tag doesn't have to match the sub-dir name exactly. As long as I configure it correctly in dockerhub, we can make the tag whatever we want. I'll merge the PR and add to dockerhub.

kapsakcj:
building on dockerhub now. Set the full dockerhub path and tag to `staphb/snippy:4.6.0-SC2` The SC part is uppercase so be mindful when adding to your nf script or elsewhere

Hope this helps!

PR #87, opened 2020-03-20
add ivar docker image

kevinlibuit:


kapsakcj:
@kevinlibuit I'm going to go ahead and merge these and push to dockerhub because I don't want to keep you or anyone else waiting.

I do have one concern about the images, but I'm not sure if it matters or not. I'll let you tell me.

HTSlib, Samtools, bwa, and iVar are all placed in the `/root` directory, which is the root user's `$HOME`. When I run the container as a non root user, with `docker run -u $(id -u):$(id -g)`, I am able to pull up help options for each program so it looks like any user can still run the programs so it shouldn't be an issue.

One problem (if it's even a problem at all?) is that if someone runs the containers as a non-root, and wants to do anything with any of the files within `/root` they will get PERMISSION DENIED since they are root owned.

I wouldn't worry about it, as long as the container is working as intended, but it's something to keep in mind if anyone encounters issues.

PR #95, opened 2020-03-31
Added cluster-report-env

AbigailShockey:
Added cluster-report-env a docker image with R installed that can be used to generate genome cluster report with R markdown


kapsakcj:
Is `cluster-report-env` OK for the dockerhub repository name?

I'll add the `cluster-report-env:latest` default tag, any other tags you would like?

AbigailShockey:
That name works. A 1.0 tag could be added as well

PR #101, opened 2020-04-30
format appropriate bed file

kevinlibuit:
Ensuring that the reference bed files all come from the .bed files rather than the .scheme.bed


kapsakcj:
will merge and rebuild on dockerhub shortly

kapsakcj:
all images have been rebuilt successfully

PR #105, opened 2020-05-14
VIBRANT

stjacqrm:
prophage identification and annotation tool


kapsakcj:
Don't worry about the build failing^ it's due to the quast container failing tests, which happens every now and then.

VIBRANT looks to be a huge docker image with its included dbs/data:
```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
kapsakcj/vibrant    latest              77f50d9f68ab        13 hours ago        11.5GB
```
Dockerhub should be able to handle this one, pending the build doesn't take too long. Going to build on dockerhub shortly

PR #115, opened 2020-05-29
fixed bwa install

erinyoung:
bwa wasn't installed correctly. It should work, now.


kapsakcj:
Is this an issue for any other ivar dockerfiles?

kapsakcj:
Looks like it might be https://github.com/StaPH-B/docker-builds/blob/71af5e40f901f3d32f131008173bdee28d1adb66/ivar/1.2.1/Dockerfile#L51

erinyoung:
Actually, now that I've looked it over, there's a bit more that needs to be done.

PR #116, opened 2020-06-01
Fixing bwa and updating the ivar + artic primers

erinyoung:
BWA should now work in these containers. Plus, a new container with the latest version of the artic primer bedfiles is included.

As bonuses, a GFF file for the reference is now included as is a bedfile for each amplicon of the artic primers.


kapsakcj:
@kevinlibuit Could you please take a look at the changes that Erin suggested in your original dockerfiles? I'd like to get your 👍 or 👎 on the changes before we merge the PR and re-build the existing docker images:
  * `staphb/ivar:1.1` 
  * `staphb/ivar:1.2.1` 
  * `staphb/ivar:1.2.1-SC2`

@erinyoung How would you like the docker tag named for your new dockerfile? `staphb/ivar:1.2.2_artic20200528` ? Or could we keep it simple at `staphb/ivar:1.2.2`

erinyoung:
I don't know if we should keep it a simple name or not. I kind of want to keep track of what version of artic is being used, especially since they are making important changes without updating their version.

kapsakcj:
ok, I agree. We'll go with `staphb/ivar:1.2.2_artic20200528`. I think dockerhub tags can accept underscores. I'll let you know what works when we get this merged and built on dockerhub

kapsakcj:
@erinyoung The `LABEL container.version` label is very vague - my apologies. I was only using that label to keep track of the versions of dockerfiles for the same program (e.g. if we had 3 versions for ivar 1.2.2 dockerfile). It really should be named `dockerfile.version`. I'll leave it for now, but it might be good to eventually rename that label across all dockerfiles

Anyways, I added a travisCI test for your new dockerfile and will merge the PR and push to dockerhub!

PR #136, opened 2020-07-23
Add VADR recipe

andersgs:
Here is VADR, complete with SARS-CoV-2 models.

I have tested it with Wuhan following the instructions here: https://github.com/nawrockie/vadr/wiki/Coronavirus-annotation

With slight modifications:

```
docker run --rm -it -v $PWD:/data vadr:latest v-annotate.pl -r -s --nomisc --lowsimterm 2 --mxsize 2000 --mkey NC_045512 --fstlowthr 0.0 --alt_fail lowscore,fsthicnf,fstlocnf --lowsc 0.75 <fasta-file-to-annotate> <output-directory-to-create>
```


kapsakcj:
Thanks for yet another PR! Nice work on this one. We were hesitant to add VADR to our repository, since the docs state it can use up to 64GB RAM, but that shouldn't stop us.

I was able to build the docker image successfully and ran the Wuhan SC2 example with no issues in Docker. Should be fine in singularity too, but we'll need to push to dockerhub first so I can convert it to Singularity format to test.

I've made some minor changes to the dockerfile, and I'll push commits in a bit:
  * deleted an extra curly brace
  * added some `apt-get` commands to clean up apt garbage and (barely) reduce the size of the docker image

This docker image is pretty large in size (9.86 GBytes uncompressed), I believe due to the vadr models, so the more we can do to reduce the size of the image the better.

Issue #118, opened 2020-06-02
LABEL container.version --> LABEL dockerfile.version


kapsakcj:
`LABEL container.version` is super vague. It should be `LABEL dockerfile.version`. Intended for tracking multiple versions of a dockerfile for a single program (and it's single version).

 For example if I had to make adjustments to the SPAdes 3.14.1 dockerfile (for which there is only one dockerfile version currently), I would bump to `LABEL dockerfile.version=2`

erinyoung:
Could you give an example of how you want this to look?

kapsakcj:
`LABEL dockerfile.version="1"` for the first iteration of a dockerfile. Then if you had to tweak the dockerfile at a later date, but not change the version of the actual program (SPAdes or whathave you), then I would change it to `LABEL dockerfile.version="2"` when commiting the edited dockerfile to the repo.

IMO semantic versioning isn't necessary here because I don't expect dockerfiles to change much once they are running smoothly and I would like to keep it simple.

Happy to hear your thoughts on this.

erinyoung:
The label is correct in the template, right?

https://github.com/StaPH-B/docker-builds/blob/master/dockerfile-template/Dockerfile

@kapsakcj , do you think we need to add/change labels?

kapsakcj:
Yes, the template has the correct LABEL (`dockerfile.version`). 

The only slight change for the template would be something like this, if multiple people are maintainers or authors of the dockerfile:

```Dockerfile
LABEL maintainer1="Anders Goncalves da Silva"
LABEL maintainer1.email="andersgs@gmail.com"
LABEL maintainer2="Curtis Kapsak"
LABEL maintainer2.email="curtis.kapsak@theiagen.com"
```

erinyoung:
I want to ensure that each docker image has the recommended labels, so I added a check in https://github.com/StaPH-B/docker-builds/pull/499. 

I had a difficult time adding wildcards when using docker inspect to look at labels, so I'm going to recommend that the first maintainer is maintainer and the nth maintainer is maintainer$n. So the recommendation would look like this:

```Dockerfile
LABEL maintainer="Anders Goncalves da Silva"
LABEL maintainer.email="andersgs@gmail.com"
LABEL maintainer2="Curtis Kapsak"
LABEL maintainer2.email="curtis.kapsak@theiagen.com"
```

erinyoung:
I think we can close this issue now.

PR #151, opened 2020-09-24
Issue 129

tgallagh:
Made a dockerfile to build image for metaphlan3.0.3. And updated README.md and travis.yml. Note that Docker version >= 17.05 is necessary to build the image which uses multistage builds to reduce image size (https://docs.docker.com/develop/develop-images/multistage-build/), but the built image should work with any docker version. 

Also, this is my first time contributing (and hopefully won't be my last). Please let me know if there's anything I should do differently for future PRs! Thanks!!!


kapsakcj:
Hey Tara, thanks so much for the PR. And thank you for updating the readme and travis tests! Sorry it took me so long to respond - I hope this docker image isn't mission critical.

I haven't played with multi-stage builds, but have been helping Lee K. as he is working on a multi-stage build for one of his projects. So that being said, I'm not sure the best practices for optimizing multi-stage builds. And I've never used metaphlan either...I'll leave the testing up to you 😄 

I have a few requests that I list here, and for nitpicky comments, I'll make comments to the specific lines in the code.
  * Please add a line in the table in `Program_licenses.md` with the info for metaphlan
  * Consider adding a nearly identical dockerfile that does not include the metaphlan database. Generally it's not a good idea to keep large files inside a docker image and we  have seen some issues (@erinyoung could elaborate) with kraken1 & 2 images that contain the minikraken db that is 8GB. For this reason, we've added kraken dockerfiles that do not download the database, and instead we've provided instructions for users to mount a volume e.g. `docker run -v /local-kraken-db:/kraken-db-inside-container` to make the image building and downloading process smoother.

Curtis

metaphlan/3.0.3/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+COPY --from=builder /usr/bin/python3.7 /usr/bin/python3.7
+COPY --from=builder /usr/lib/python3.7 /usr/lib/python3.7
+COPY --from=builder /usr/lib/x86_64-linux-gnu/libexpat.so /usr/lib/x86_64-linux-gnu/libexpat.so
+COPY --from=builder /lib/x86_64-linux-gnu/ /lib/x86_64-linux-gnu/
+COPY --from=builder /usr/local/lib/python3.7/dist-packages/ /usr/local/lib/python3.7/dist-packages/
+COPY --from=builder /usr/lib/python3/dist-packages/* /usr/lib/python3.7/dist-packages/
+
+WORKDIR /data
+
+ENV PATH="$PATH:/usr/bin/bowtie2/bowtie2-2.3.3.1-linux-x86_64"
```
kapsakcj, line 60: Please edit/add this line to ensure compatibility with singularity:
```Dockerfile
ENV PATH="$PATH:/usr/bin/bowtie2/bowtie2-2.3.3.1-linux-x86_64"\
    LC_ALL=C
```
Probably unnecessary, but maybe some perl needs to run somewhere that isn't obvious

tests/metaphlan.sh
```diff
@@ -0,0 +1,5 @@
+#/bin/bash
+# test for spades container
```
kapsakcj, line 2: `# test for metaphlan container`

metaphlan/3.0.3/Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM ubuntu:bionic AS builder
+
+LABEL base.image="ubuntu:bionic"
+LABEL version="1"
```
kapsakcj, line 4: This is an old LABEL that is pretty vague and useless. Please change to this:
`LABEL dockerfile.version="1"`

If we need to adjust the dockerfile (after merging PR into master, building image on dockerhub) then we should increment this to 2. Semantic versioning is overkill here, and I just wanted a way to track dockerfile versions in case we need to tweak them for a specific version of a program. More info here https://github.com/StaPH-B/docker-builds/issues/118

metaphlan/3.0.3/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+LABEL software="MetaPhlAn3.0"
+LABEL software.version="3.0"
```
kapsakcj, line 6: `LABEL software.version=3.0.3` ?

tests/metaphlan.sh
```diff
@@ -0,0 +1,5 @@
+#/bin/bash
```
kapsakcj, line 1: `#!/bin/bash`
Missing the '!'

metaphlan/3.0.3/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+LABEL description="microbial composition of metagenomes"
+LABEL website="http://cab.spbu.ru/files/release3.13.0/manual.html"
```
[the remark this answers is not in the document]
tgallagh: Thanks, fixed

metaphlan/3.0.3/Dockerfile
tgallagh: Thanks, I added the env path to dockerfiles.

For the perl comment, I'm assuming this is just a simple check to make sure locale settings are in the image. I added an empty print line (RUN perl -e 'print') in each dockerfile (for 3.0.3 = line 87 and 3.0.3-no-db = line 91). Is this what you meant? Let me know if you want me to change

tests/metaphlan.sh
tgallagh: Fixed, thanks!

metaphlan/3.0.3/Dockerfile
tgallagh: Thanks, I updated the dockerfile.version labels in both dockerfiles. 

I also moved the labels down in the Dockerfile, so that the labels are associated with the final docker image that's generated from the multistage build

metaphlan/3.0.3/Dockerfile
tgallagh: Fixed, thanks

tests/metaphlan.sh
tgallagh: Thanks, fixed

tgallagh:
Hi Curtis,

Thank you again for your feedback and help with this! Sorry the long response - I've been doing a lot of wetlab lately. I don't need to use the metaphlan docker anytime soon, so no need to rush for reviewing. 😄 

As requested, I updated the program licenses file. I also made two versions of the dockerfile: one that builds the ~3GB metaphlan database (3.0.3) and one without (3.0.3-no-db). For 3.0.3.-no-db, I included instructions on how to install the metaphlan database in metaphlan/3.0.3-no-db/README.md. 

I recently realized that the multistage builds do not automatically delete or overwrite intermediate stages, so the builder will need to delete any intermediate stages after the final image build is complete. I labeled the intermediate stages so they can be easily deleted and noted this in the Dockerfile comments:

For 3.0.3:
docker image ls --filter "label=stage=builder_metaphlan"
docker image prune --filter "label=stage=builder_metaphlan"

For 3.0.3-no-db:
docker image ls --filter "label=stage=builder_metaphlan_nodb" 
docker image prune --filter "label=stage=builder_metaphlan_nodb"

Sorry about that - let me know if you want me to avoid multistage builds in the future!

I tested both versions in docker, still need to test in singularity. 

thanks again! 
Tara

Program_Licenses.md
```diff
@@ -37,6 +37,7 @@ The licenses of the open-source software that is contained in these Docker image
 | Mash | non-standard license (see link) | https://github.com/marbl/Mash/blob/master/LICENSE.txt |
 | mashtree | GNU GPLv3 | https://github.com/lskatz/mashtree/blob/master/LICENSE |
 | Medaka | Mozilla Public License 2.0 | https://github.com/nanoporetech/medaka/blob/master/LICENSE.md |
+| Metaphlan | non-standard license (see link) | https://github.com/biobakery/MetaPhlAn/blob/3.0/license.txt |
```
kapsakcj, line 40: I believe they do use the standard MIT license

metaphlan/3.0.3/Dockerfile
kapsakcj: Yeah, for some reason when docker images are converted to singularity format, locale settings are not saved and perl programs run into locale settings errors. Doesn't happen with docker, only singularity. That's why I typically will set LC_ALL=C to make sure its baked into the image

RE lines 87 and 91 in dockerfiles - I didn't realize you could check it that way, that helps! Only downside is that for the no-db dockerfile, line 91 adds an additional layer if you give it its own RUN statement. Perhaps combine `perl -e 'print'` test with the previous layer?
```Dockerfile
# make dir for metaphlan database
RUN mkdir /usr/local/lib/python3.7/dist-packages/metaphlan/metaphlan_databases && \
 ln -s /usr/bin/python3.7 /usr/bin/python && \
 perl -e 'print'
```

Program_Licenses.md
tgallagh: you're right, thanks for catching that. I'll fix it

metaphlan/3.0.3/Dockerfile
tgallagh: Actually, I'm not sure if perl will always generate a warning if it's unable to find locale settings. I need to look further into this. Since this is still new to me 😅 , I'll change to whatever you typically do in regards to your 1st comment - "Probably unnecessary, but maybe some perl needs to run somewhere that isn't obvious". Do you have an example in one of your dockerfiles? I'll also fix the no-db dockerfile and get rid of running that extra layer. 

thanks!

kapsakcj: it's an annoying ~~error~~ warning that looks like this. I pulled an older, non-staphb prokka docker image (https://hub.docker.com/r/hadrieng/prokka/dockerfile) with singularity, entered singularity shell, and pulled up help options.
```
Singularity> prokka --help
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
        LANGUAGE = (unset),
        LC_ALL = (unset),
        LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
Name:
  Prokka 1.12 by Torsten Seemann <torsten.seemann@gmail.com>
```

kapsakcj:
Thanks for the additional dockerfile, the readme, and for making those changes.

If you think multi-stage build is the way to go, I say go for it. I just have less/almost no experience with multi-stage so it's a little harder for me to test & debug.

If you're happy with these dockerfiles, we can merge the PR, and I can set up the builds in dockerhub. Once they're built, we can download/convert them with singularity and test.

Curtis

tgallagh:
Thanks Curtis. Before you merge, I'll incorp those 2 changes and I also want to test in singularity. Will be in touch soon!

tgallagh:
Hi Curtis,

Thanks for waiting, I think it's ready! 

Quick update - I think I misunderstood your first request about running perl to check locale settings in the env, so I removed those perl lines from each dockerfile (they didn't do anything to check LC settings during the build process). I tested both builds in singularity and added examples to each README. 

Let me know what else I can do. And please let me know if there's anything I should do differently for future Dockerfiles and PRs.

Thanks,
Tara

kapsakcj:
Awesome work. I think this is ready to be merged. Thank you so much for the contribution.

I will set up the builds on dockerhub using the name and tags you have specified in the READMEs

PR #159, opened 2020-12-21
Update pangolin to version 2.1.1 and pangoLEARN to 2020-12-17

andersgs:
These version cover the B.1.1.7 lineage from the UK


kapsakcj:
Hey Anders - thanks for the PR. This dockerfile builds fine on my machine and I can pull up `pangolin` help options and 2.1.1 version. I'm going to merge and build on dockerhub so it's available for folks to use.

Please let us know if you have any issues using it down the road and we can tweak the dockerfile. I'll comment back here when it's built on dockerhub.

PR #167, opened 2021-01-13
Update AMRFinderPlus 3.9.8

andersgs:


kapsakcj:
Hi Anders, thanks for another PR. The docker image builds and runs successfully, but I want to suggest some changes to help improve it.

- switch to `ubuntu:focal` for base image
- install ncbi-blast+ and hmmer via `apt-get` instead of manually. More recent versions of `ncbi-blast+` (v2.9.0) and `hmmer` (v3.3) can be installed via apt-get with a more recent version of Ubuntu

These changes significantly reduce the size of the docker image, simplifies the dockerfile, and speeds up the image build process

Saves almost half of a GB:
```
anders/amrfinder                            3.9.8-focal                   578aa37789b6   15 minutes ago   616MB
anders/amrfinder                            3.9.8-bionic                  32321266eb64   19 minutes ago   1.22GB
```

I'm going to push a commit with these changes, but please let me know if you'd like to keep the dockerfile the way you have it. We can revert the commit.

kapsakcj:
With my changes to the dockerfile, the docker image still builds successfully and I ran one E. coli assembly through:
```
$ amrfinder --threads 8 -o amrfinder-3.9.8-focal-cleaned-up-test -n ecoli.skesa.fasta --plus -O Escherichia
Running: amrfinder --threads 8 -o amrfinder-3.9.8-focal-cleaned-up-test -n ecoli.skesa.fasta --plus -O Escherichia
Software directory: '/amrfinder/'
Software version: 3.9.8
Database directory: '/amrfinder/data/2020-12-17.1'
Database version: 2020-12-17.1
AMRFinder translated nucleotide and mutation search
Running blastx...
Running blastn...
AMRFinder took 65 seconds to complete
```

Let me know what you'd like to do!

andersgs:
@kapsakcj that is a great change! 

I was going to make a few changes to try to minimise the size too, such as multistage build and organise the location of the installed files all to a single folder, rather than directly on root (/) as they are now.

but, i need image urgently...

andersgs:
My only concern is that the BLAST version might be too old... And, v2.9.0 may just be on the cusp of acceptable, as the blast database versions change from 4 to 5, and 2.9.0 might be the last version that supports both? I think...

kapsakcj:
Got it. How about we move to ubuntu:groovy? It has ncbi-blast+ v2.10.1 available via apt

kapsakcj:
I can test out with ubuntu:groovy as the new base image

kapsakcj:
OK, with the new base image, everything builds fine and runs as expected too. I will merge the PR and build on dockerhub

andersgs:
Thank you @kapsakcj!!

kapsakcj:
No problem and thank you for the dockerfile!

The new docker image is now available with the new tag, `staphb/ncbi-amrfinderplus:3.9.8`, and `staphb/ncbi-amrfinderplus:latest` tag (until we add a new version)

It is way smaller than I expected - uncompressed 650MB and compressed 178MB.

Please let me know if you run into any issues!

PR #171, opened 2021-01-22
Fixed artic dockerfiles and added to readme

AbigailShockey:


kapsakcj:
Thanks for the updates! We ran out of free TravisCI credits so that check won't run or complete. I'll merge and rebuild on dockerhub shortly

PR #174, opened 2021-02-05
Update pangolin to v2.1.11

andersgs:


kapsakcj:
Thanks for the PR. I'll push to dockerhub.

FYI I added a couple of things in the last dockerfile, pangolin v2.1.10, to try to reduce the size of the image. It helped a tiny bit, but the image is still pretty large. I don't know if we can get around it due to the number of dependencies for pangolin. 

https://github.com/StaPH-B/docker-builds/blob/fe60b3fdfb94f7e1327bef9d912c6ec63a9a5125/pangolin/2.1.10/Dockerfile#L27

and a bit further down:
https://github.com/StaPH-B/docker-builds/blob/fe60b3fdfb94f7e1327bef9d912c6ec63a9a5125/pangolin/2.1.10/Dockerfile#L72

PR #178, opened 2021-02-19
Add Treemmer

AbigailShockey:


kapsakcj:
Hi Abigail - thanks for the PR. Is there any particular reason for the separate `pip2 install X` commands? The docker image might build faster & more efficiently if you combine that into one command e.g. `pip2 install numpy==1.16.6 ete3==3.1.1 <others>`.

Another suggested change - could you combine lines 39 and 40 into one layer?
```Dockerfile
ENV PATH="/Treemmer:${PATH}" \
  LC_ALL=C
```

And I think you are ok with line 33, but it would be safer to include a `/` in front of data e.g. `mkdir /data` to ensure that it ends up at the root of the filesystem

Thanks for the PR, this is the first I've heard of Treemmer!

AbigailShockey:
Fixed issues in previous comment

kapsakcj:
Thanks, those changes look good.

Also.....sorry for being a bother, but I noticed this on the gitlab site
> Treemmer moved to GitHub and this page is not updated: please find the last version of Treemmer here : https://github.com/fmenardo/Treemmer

I think it might be better to list the github urls instead of gitlab

kapsakcj:
ok, its available on dockerhub now

Issue #163, opened 2021-01-07
automated testing idea: include SARS-CoV2 genomes/fastas


kapsakcj:
I think it would be a good idea to have a couple of SARS-CoV2 assemblies/consensus sequences available in the repo to aid in automated testing of COVID19 related tools. Tiny genome = tiny fasta = ideal for automated tests.

Perhaps the Hu-1 strain, maybe a B.1.1.7, other important ones?

could test these tools:
  * pangolin
  * VADR
  * iVar (? not sure on this one)
  * others?

e.g. 
`/tests/SC2-test-genomes.fasta` would have something along the lines of...
```
>Hu-1
ATCGAGC....
>B.1.1.7
CTATCGA...
```

First use TravisCI to build the docker image (`docker build -t pangolin:2.1.6 pangolin/2.1.6`) & bash scripts to test it
`/tests/pangolin.sh`
```bash
#!/bin/bash
# test for pangolin
set -e

docker run pangolin:2.1.6 pangolin --threads 2 --outdir pangolin-test /tests/SC2-test-genomes.fasta
```

kapsakcj:
VIBRANT too?

kapsakcj:
ok, maybe drop VADR from the list. It requires a lot of RAM to run and that might not be feasible on a TravisCI VM.

erinyoung:
I have some bam files, fastq, and fastas that I could upload. Where do you want me to put them?

kapsakcj:
Great. I think either directly in `/tests` or a sub-folder within tests would be fine.

Side note: We have unintentionally abused TravisCI to the point where the StaPH-B organization has run out of "free credits" for automated tests. Partly because it would run tests every time a commit was pushed to `master` branch; partly because our automated tests would take a LONG time to run (anywhere from 15-50 minutes) which ate through all of our free credits

Now may be a good time to look into GitHub Actions unless we want to start forking out some money for the service 😬 

So I appreciate you taking the initiative to adding these files, but just be aware that TravisCI tests won't run

@k-florek @kevinlibuit FYI - this impacts staphb_toolkit too unfortunately ☹️

k-florek:
[...]

PR #184, opened 2021-03-12
Migrate Travis workflow to GitHub Action

fanninpm:
This GitHub Action would probably not run as often as the old Travis workflow.


fanninpm:
I'm sure there could be more to do with testing proper version numbers, but it's at least a start.

fanninpm:
Let me know if I need to squash all those commits.

kapsakcj:
Hi @fanninpm , thank you for this awesome PR! You must have seen issue #163 . I've been meaning to switch to GitHub actions over Travis due to the limits on Travis without paying $$. So thank you!

I'd like to get some other folk's eyes on this before we merge and I'll need some time to review as well. I'm new to GitHub actions so after a quick scan, I couldn't exactly tell what was going on.

@k-florek @erinyoung interested in reviewing?

fanninpm:
@kapsakcj I'm also not that experienced with GitHub Actions, although I found [this documentation page](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) to be extremely helpful for writing the action. (If you find that it's rather dense and confusing, then I think you could start at the [Quickstart guide](https://docs.github.com/en/actions/quickstart).) If you want, I can walk through the code and show you what everything does.

fanninpm:
The following is my explanation of the new GitHub Actions YAML file:

- `name: Build and Test` — This is mostly for cosmetic purposes, and its value is displayed in places such as the [Actions tab](https://github.com/StaPH-B/docker-builds/actions) (at the top of the page) or the page of a PR (e.g. towards the bottom of the page). If this parameter is omitted, then the value defaults to something like ".github/workflows/build-and-test.yml".
- `on: [push, pull_request]` — This required section tells GitHub if and when the workflow should be run. Currently, this runs anytime this workflow is present and a commit is either pushed to any branch or a commit is pushed in reference to a PR. This can (and, in my opinion, should) be further refined, [as explained in the docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on).
- `jobs:` — This tells GitHub what exactly it should be doing. This workflow has two jobs specified, explained below:
  - `generate-matrix:` — This job generates a matrix for the next job to use. It has three `steps`, which are run sequentially:
    - `uses: actions/checkout@v2` — This simply checks out the repository. Nothing special so far.
    - `name: Check changed files` (`id: diff`) — This finds out which files changed [(using `git diff`)](https://github.community/t/check-pushed-file-changes-with-git-diff-tree-in-github-actions/17220/10), and emits those as output for the next step.
    - `name: Set matrix for build` (`id: set-matrix`) — This takes the output of the previous (`diff`) step and [emits a JSON object](https://stackoverflow.com/a/62953566/11948346) consisting of the names and versions of the images we would like to test, as defined in the `image_versions` function. I set the `image_versions` function up such that it can be expanded with more images (by adding new cases to the `case...esac` block), as well as with multiple versions per image (by modifying the `ref_versions` array inside that image's case). Shell scripts with the same name as the image can be added to the `tests/` directory to test.
  - `build-and-test:` — This job specification generates a job for each image/version pair in the matrix generated by `generate-matrix`. If no jobs are in the matrix, then this job is skipped, making the workflow exit cleanly. Otherwise, it sets up a batch of jobs to be run in parallel, building and testing a single version of an image in the matrix. All of these must pass in order for the workflow to exit cleanly.

kapsakcj:
Hey @fanninpm I truly appreciate the explanation. That helps a lot.

Apologies for the delay - I've recently changed jobs and am preparing to move to a new home so this has fallen lower on my priority list.

That said - I haven't forgotten about the PR and will spend time reviewing when I'm able. I really like what you've done and I foresee that we could build upon this GH Actions workflow for testing many of the various containers. 

Thank you!

kapsakcj:
Hi @fanninpm , finally getting around to reviewing this. Thank you for your patience

I would like to see the output/logs from these tests and unfortunately GitHub removes them after X amount of time so I can't see the logs from the past history on your forked repo (6 months old at this point).

Would it be OK if I push a simple commit to this PR which should trigger GH actions to run?

Thinking of just removing the "build passing" badge on the README since we're getting rid of Travis anyways.

fanninpm:
It'll be a couple of hours before I can get to this and rebase my branch. If you want, you can push that commit on your own time.

erinyoung:
First of all, I want to thank you for your efforts. I've gone through your files multiple times and they have helped me learn about github actions. I think we're going to go with something simpler moving forward.

I already commented on another PR that you were involved in (https://github.com/StaPH-B/docker-builds/pull/244#issuecomment-1043354987).

Basically, tests will be in the dockerfile itself instead of being in a separate script. I think this is better because the creator can test their 'test' by using `docker build --target test` as they create their test. Thus, the creator only needs to create their Dockerfile, README, and test-tool.yml file (plus updating the main License and README). This also simplifies the build*yml files.

I do appreciate your tests for spades, mash, and mashtree, but I don't think we can merge this PR because we're moving in another direction.

PR #190, opened 2021-03-23
Erin dev

erinyoung:
I found a some SARS-CoV-2 sequencing files that are small enough to upload to github. 

These are intended to be test files as mentioned in [issue 163](https://github.com/StaPH-B/docker-builds/issues/163)


kapsakcj:
Thanks for the PR with these test files. This will be valuable for testing various software/docker images.

I think it would be useful to list as much metadata as possible about this test data. Some metadata is captured in NCBI SRA but it would be helpful to have it here too.

Do you know what nextclade & PANGO lineage this sample is expected to be? 

Not sure if it is useful or not, but maybe list the library prep & sequencing kit/chemistry in the README as well? 

Ideas on what other metadata would be useful?

kapsakcj:
Was the consensus fasta generated using Cecret? What version of Cecret?

erinyoung:
Admittedly, if the the library prep and chemistry info isn't in the SRA metadata, I don't know it.

I've added the workflow version number and tool version numbers.

kapsakcj:
Got it. Thanks for making those changes. I'm going to go ahead and merge and we can incorporate automated tests later in a separate PR, preferably after I review & merge PR #184  so that we are using GitHub actions and not Travis. Would be good to leverage the automated testing framework that is in #184

PR #196, opened 2021-04-15
adding gnuplot

erinyoung:
plot-ampliconstats requires gnuplot


kapsakcj:
Thanks for the fix, Erin! I'll rebuild the image on dockerhub momentarily

PR #205, opened 2021-05-19
creating ivar:1.3.1_Titan container

jvhagey:


kapsakcj:
I'm going to merge this into master and build on dockerhub as `staphb/ivar:1.3.1-titan`

I did a quick spot check on one SC2 sample and everything ran as expected. More testing would be welcome.

The only other thing to maybe add is minimap2, if folks request it. Can do that later and re-build the images if folks request it.

PR #207, opened 2021-06-02
Added Kraken2 inspect txt file

jvhagey:


jvhagey:
Extra commits need to be removed

PR #217, opened 2021-06-28
added dockerfile for pangolin 3.1.5 & associated dependencies

kapsakcj:
hoping this dockerfile won't be out of date in a day or so....


kapsakcj:
Builds with no issues and runs successfully on a handful of fastas in both default pangolearn and usher modes.

building on dockerhub first and then syncing with quay

PR #224, opened 2021-07-08
add TSV to VCF script to iVar container

rpetit3:
The PR adds a script, [ivar_variants_to_vcf.py](https://github.com/nf-core/viralrecon/blob/dev/bin/ivar_variants_to_vcf.py), to convert the iVar variants in TSV format to VCF. The script is pulled from [nf-core/viralrecon](https://github.com/nf-core/viralrecon) v2.1 release.

`ivar_variants_to_vcf.py` had `/user/bin/env python` in it's shebang line, the container has `python3` in the path but not `python`. Rather than editing `ivar_variants_to_vcf.py`, I used `update-alternatives` to add `python` (just a symlink to `python3`).

Please let me know if there are any questions, or anything needed on my end.

### Example Usage
```
docker run --rm -u $(id -u):$(id -g) -v ${PWD}:/data ivar:1.3.1 ivar_variants_to_vcf.py 20154160_S5_L001.variants.tsv 20154160_S5_L001.variants.vcf
```

Attaching inputs/outputs (w/ .txt extension to make GitHub happy)
[20154160_S5_L001.variants.vcf.txt](https://github.com/StaPH-B/docker-builds/files/6785952/20154160_S5_L001.variants.vcf.txt)
[20154160_S5_L001.variants.tsv.txt](https://github.com/StaPH-B/docker-builds/files/6785953/20154160_S5_L001.variants.tsv.txt)

### DIFF
--- a/ivar/1.3.1/Dockerfile
+++ b/ivar/1.3.1/Dockerfile
@@ -4,6 +4,7 @@ FROM ubuntu:18.04
 ARG SAMTOOLSVER=1.12
 ARG HTSLIBVER=1.12
 ARG IVARVER=1.3.1
+ARG VIRALRECON=2.1
 
 LABEL base.image="ubuntu:18.04"
 LABEL dockerfile.version="2"
@@ -70,6 +71,15 @@ RUN cd root/ && \
   make && \
   make install && \
   mkdir /data
+  
+# nf-core/viralrecon
+RUN cd /tmp && \
+  wget https://github.com/nf-core/viralrecon/archive/refs/tags/${VIRALRECON}.tar.gz && \
+  tar -xzvf ${VIRALRECON}.tar.gz && \
+  rm -rf ${VIRALRECON}.tar.gz && \
+  chmod 755 viralrecon-${VIRALRECON}/bin/ivar_variants_to_vcf.py && \
+  cp viralrecon-${VIRALRECON}/bin/ivar_variants_to_vcf.py /usr/local/bin && \
+  rm -rf viralrecon-${VIRALRECON}/
 
 # set /data as working directory
 WORKDIR /data
--- a/ivar/1.3.1/Dockerfile
+++ b/ivar/1.3.1/Dockerfile
@@ -7,7 +7,7 @@ ARG IVARVER=1.3.1
 ARG VIRALRECON=2.1
 
 LABEL base.image="ubuntu:18.04"
-LABEL dockerfile.version="2"
+LABEL dockerfile.version="3"
 LABEL software="iVar"
 LABEL software.version="1.3.1"
 LABEL description="Computational package that contains functions broadly useful for viral amplicon-based sequencing."
--- a/ivar/1.3.1/Dockerfile
+++ b/ivar/1.3.1/Dockerfile
@@ -79,7 +79,8 @@ RUN cd /tmp && \
   rm -rf ${VIRALRECON}.tar.gz && \
   chmod 755 viralrecon-${VIRALRECON}/bin/ivar_variants_to_vcf.py && \
   cp viralrecon-${VIRALRECON}/bin/ivar_variants_to_vcf.py /usr/local/bin && \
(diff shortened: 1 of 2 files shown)


kapsakcj:
Apologies I didn't get to this sooner - just have one quick question

Will updating the python symlink to python3 negatively affect any users of this container? I'm trying to think of a scenario where that is possible but most if not all tools included in the container here are not python-based. 

Only scenario I can think of is if someone is using the container to run python based code (whether included in the container or added by user at runtime), but if so, they should probably be using a different container.... so 🤷

rpetit3:
Totally agree, looking at the current ivar 1.3.1 container, it looks like only Python3 is installed:

```
docker run --rm -u $(id -u):$(id -g) -v ${PWD}:/data staphb/ivar:1.3.1 ls -lh /usr/bin/ | grep python
lrwxrwxrwx 1 root root      26 Mar 26  2018 dh_pypy -> ../share/dh-python/dh_pypy
lrwxrwxrwx 1 root root      29 Mar 26  2018 dh_python3 -> ../share/dh-python/dh_python3
lrwxrwxrwx 1 root root      23 Jan 26 15:33 pdb3.6 -> ../lib/python3.6/pdb.py
lrwxrwxrwx 1 root root      31 Oct 25  2018 py3versions -> ../share/python3/py3versions.py
lrwxrwxrwx 1 root root      26 Mar 26  2018 pybuild -> ../share/dh-python/pybuild
lrwxrwxrwx 1 root root       9 Oct 25  2018 python3 -> python3.6
lrwxrwxrwx 1 root root      16 Oct 25  2018 python3-config -> python3.6-config
-rwxr-xr-x 2 root root    4.4M Jan 26 15:33 python3.6
lrwxrwxrwx 1 root root      33 Jan 26 15:33 python3.6-config -> x86_64-linux-gnu-python3.6-config
-rwxr-xr-x 2 root root    4.4M Jan 26 15:33 python3.6m
lrwxrwxrwx 1 root root      34 Jan 26 15:33 python3.6m-config -> x86_64-linux-gnu-python3.6m-config
lrwxrwxrwx 1 root root      10 Oct 25  2018 python3m -> python3.6m
lrwxrwxrwx 1 root root      17 Oct 25  2018 python3m-config -> python3.6m-config
lrwxrwxrwx 1 root root      33 Oct 25  2018 x86_64-linux-gnu-python3-config -> x86_64-linux-gnu-python3.6-config
lrwxrwxrwx 1 root root      34 Jan 26 15:33 x86_64-linux-gnu-python3.6-config -> x86_64-linux-gnu-python3.6m-config
-rwxr-xr-x 1 root root    3.3K Jan 26 15:33 x86_64-linux-gnu-python3.6m-config
lrwxrwxrwx 1 root root      34 Oct 25  2018 x86_64-linux-gnu-python3m-config -> x86_64-linux-gnu-python3.6m-config
```

As an alternative, I could just change the shebang line to explicitly use Python3 (`/usr/bin/env python3` instead of the current `/usr/bin/env python`). 

I think that might be a better approach since its not messing with global python settings.

kapsakcj:
I think it's OK to leave as is, no need to change the shebang line of the script.

I'm just a little nervous because the ivar container is the most heavily downloaded/used container in our collection.

Yes, it's a global change to python settings, but really it does not change any underlying install of python3 and python2 wasn't there to begin with.

We should be OK. So I propose this, as an attempt to do a little bit more testing before we rebuild the `staphb/ivar:1.3.1` docker image tag.

- merge this PR
- build on dockerhub, but with a different tag, e.g. `staphb/ivar:1.3.1-vcf-script`
- run some tests on Terra with `staphb/ivar:1.3.1-vcf-script`, confirm that Titan PE, particularly the iVar steps, are not affected by this change
- if all good - rebuild the `staphb/ivar:1.3.1` tag with this dockerfile 
- if not good - find another solution

What do ya think? Would you be able to test out on your end too?

rpetit3:
closing in favor of https://github.com/StaPH-B/docker-builds/pull/245

PR #244, opened 2021-08-02
GitHub actions testing

lskatz:
Hi, I would like to contribute docker build tests into the repo.  There are two steps in a test:

1. build the image
2. load the image and run a test command (usually `executable --help`)

I have a pretty good yaml ironed out for FastANI and Mashtree, and the only difference is some variable values.  It should be easy to copy over to include all the other builds.  The variables are at the top of the yaml and look like this:

```
env:
  CONTAINER_NAME: mashtree
  VERSION: "1.2.0"
  # The inspect command will be run inside the container
  # with `docker run` and should have an exit code of 0
  # or an exit code > 0 if the container is somehow bad.
  INSPECT_CMD: "mashtree -h"
```

The `CONTAINER_NAME` reflects the subfolder in this repo.  The `VERSION` reflects the sub-subfolder.  `INSPECT_CMD` is some proxy command that will let you know if the container runs properly.  For something more in-depth, you can use inherent software checks.  In this case, it would be something like `perl Makefile.PL && make && make test`.  SPAdes has a `test` subcommand too, for another example.

Here is an example of a passing test for Mashtree: https://github.com/lskatz/docker-builds/actions/runs/1091164975 

Caveat: I haven't investigated if this counts against any quota.


lskatz:
PS: thank the git gods that there is "squash and merge" because there were a lot of debugging steps in here

fanninpm:
See also #184

lskatz:
Amazing.  I think there is a good set of ideas here that should be combined.  For example I didn't think of a yaml list to just put all the tests into one script and that would be great.

lskatz:
@fanninpm would you mind if I tried to merge some of your ideas with mine in my pull request?

fanninpm:
Go ahead. It may take a while for @kapsakcj to get around to it, though.

You might want to take a look at [the tests directory,](https://github.com/StaPH-B/docker-builds/tree/master/tests) which has a bunch of shell scripts. It might be better to have one GitHub action YML file that can test all the packages (or only the packages that changed) instead of one file that only tests one package. My implementation might run into a problem with scale, as [GitHub sets a limit for the number of jobs that can be in a job matrix.](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits)

lskatz:
Ok great.  I'll look into that option and expanding the tests directory.  The URL you referenced fortunately says "If [concurrent jobs] exceeded, any additional jobs are queued" and so I guess it just means we can only do 20 or 40 at a time but they will all eventually finish.  If I'm reading it right.

fanninpm:
I was actually referring to the bullet point below that:

> **Job matrix** - A job matrix can generate a maximum of 256 jobs per workflow run. This limit also applies to self-hosted runners.

Broadly speaking, here's how I set up the workflow run:
- The `generate-matrix` matrix looks at which Dockerfiles have been modified since a given commit and constructs a JSON of name-version pairs.
- The `build-and-test` matrix spawns one job for each name-version pair that builds its Docker image and tests it. (This is the part that not well-guarded against massive changes.)

erinyoung:
Thank you for your pull request! Github actions are new to me, but your efforts with those of @SarahNadeau have set up something different for this repo and I think will be simpler and easier to maintain.

I think the direction we're going to be going with is to have a test portion in your dockerfile combined with a test.yaml in '.github/workflows/test-tool.yml'

Example htslib Dockerfile:
```
ARG HTSLIBVER="1.14"

FROM ubuntu:xenial as app

ARG HTSLIBVER

LABEL base.image="ubuntu:xenial"
LABEL dockerfile.version="1"
LABEL software="htslib"
LABEL software.version="1.14"
LABEL description="A C library for reading/writing high-throughput sequencing data"
LABEL website="https://github.com/samtools/htslib"
LABEL license="https://github.com/samtools/htslib/blob/develop/LICENSE"
LABEL maintainer="Erin Young"
LABEL maintainer.email="eriny@utah.gov"

# install dependencies, cleanup apt garbage
# It's helpful when they're all listed on https://github.com/samtools/htslib/blob/develop/INSTALL
RUN apt-get update && apt-get install --no-install-recommends -y \
 wget \
 ca-certificates \
 make \
 bzip2 \
 autoconf \
 automake \
 make \
 gcc \
 perl \
 zlib1g-dev \
 libbz2-dev \
 liblzma-dev \
 libcurl4-gnutls-dev \
 libssl-dev && \
 rm -rf /var/lib/apt/lists/* && apt-get autoclean

# get htslib and make /data
RUN wget https://github.com/samtools/htslib/releases/download/${HTSLIBVER}/htslib-${HTSLIBVER}.tar.bz2 && \
 tar -vxjf htslib-${HTSLIBVER}.tar.bz2 && \
 rm htslib-${HTSLIBVER}.tar.bz2 && \
 cd htslib-${HTSLIBVER} && \
 make && \
 make install && \
 mkdir /data

# set $PATH (honestly unnecessary here, lol) and locale settings for singularity compatibility
ENV PATH="$PATH" \
 LC_ALL=C

# set working directory
WORKDIR /data

# testing
FROM app as test

ARG HTSLIBVER

WORKDIR /htslib-${HTSLIBVER}/test

RUN perl /htslib-${HTSLIBVER}/test/test.pl
```
Because bash commands are easier for me to follow and it's easy for the creator to test their image prior to their pull request by using the `--target test` option.

```
docker build --target test --tag htslib:test docker-builds/htslib/1.14
```

And then the corresponding yaml file that triggers the github action either manually or automatically with a pull request.
.github/workflows/test-htslib.yml
```
name: Test htslib image

on:
  workflow_dispatch:
  pull_request:
    paths:
      - 'htslib/1.14/Dockerfile'

jobs:
  build-to-test:
    uses: ./.github/workflows/build-to-test.yml
    with:
      path_to_context: "htslib/1.14"
      dockerfile_name: "Dockerfile"
      cache: "htslib"
```

erinyoung:
Since your current yaml for testing mashtree is a very different format, I was thinking of closing this pull request and waiting for you to open a new one.

PR #302, opened 2022-02-24
fix typo in workflow input repository name

SarahNadeau:
- [x] This comment contains a description of what is in the pull request.

This PR is to fix a typo in an input variable for two of the called GitHub actions workflows. 

- [x] Update relevant GitHub actions workflow files

Changed "repositoy_name" to "repository_name". This variable has not been used in any caller workflows because it has a default ("staphb"), so it only has to be fixed in these 2 files. I'd like to fix it because I want to use it to keep deploying to my own Docker Hub from my fork of this repo.

NA.

NA.

### DIFF
--- a/.github/workflows/build-to-deploy.yml
+++ b/.github/workflows/build-to-deploy.yml
@@ -17,7 +17,7 @@ on:
         default: Dockerfile
         required: false
         type: string
-      repositoy_name:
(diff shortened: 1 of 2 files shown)


k-florek:
Thanks for catching this!

PR #315, opened 2022-03-15
Erin samtoolshtslibbcftools

erinyoung:
- [ X] This comment contains a description of what is in the pull request.

This PR is to update samtools, bcftools, and htslib to the current versions.

There's also a duplicate bonus of having masurca in here. (If I need to take it out I can.)

- [ X] Build your own docker image using a Dockerfile
  - [ X] Directory structure should be name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. spades/3.12.0/Dockerfile)
  - [ X] Includes the recommended LABELS
- [ X] (Optional) Dockerfile is built with best practices and has been approved by a linter (such as https://hadolint.github.io/hadolint/)
- [ X] Edit main README.md
- [ X] Ensure tool is listed in Program_Licenses.md
- [ X] Create a simple container-specific README.md in the same directory as the Dockerfile (i.e. spades/3.12.0/README.md)
- [ X] Update GitHub actions workflow if needed
- [ X] Any files required for building are located in the same directory as the Dockerfile (i.e. spades/3.12.0/my_spades_tests.sh)
- [ X] Have successfully run the workflow "Test <program name> image" in your forked repository


erinyoung:
Now samtools has python3!

kapsakcj:
Really awesome work on this. I'm excited to upgrade to the latest versions of things. I'll cover my recommendations when we meet tomorrow, but wanted to jot one down now.

Avoid using `ubuntu:impish` as a base image. I did the same with `ubuntu:groovy` until I realized that the company stopped supporting that release of Ubuntu since it's an "interim release" with a short time period of support. 

While the base images should exist on dockerhub for a long time, they will no longer be updated and receive security updates. `ubuntu:groovy` on dockerhub has not been updated in 8 months

I would either go with `ubuntu:focal` or if you want more bleeding edge/recent release then use `ubuntu:jammy` 

https://ubuntu.com/about/release-cycle

erinyoung:
My apologies for slacking on this. All the images have changed to focal, and I've updated the deploy yml files.

htslib/1.15/Dockerfile
```diff
@@ -0,0 +1,60 @@
+# for easy upgrade later. ARG variables only persist during build time
+ARG HTSLIBVER="1.15"
```
kapsakcj, line 2: I'm being nitpicky here, but I think we can overwrite line 6 with line 2 instead.

So line 6 will be `ARG HTSLIBVER="1.15"` and we delete lines 1-3.

kapsakcj: Disregard, I should have read the docs: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

kapsakcj (changes requested):
All looks great to me. We can test the 3 deploy workflows once the PR is merged.

Just would like to request the one small change to the samtools 1.15 dockerfile. I'm happy to push the commit if that is ok with you.

samtools/1.15/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+LABEL dockerfile.version="1"
+LABEL software="Samtools"
+LABEL software.version="${SAMTOOLSVER}"
+LABEL description="Tools (written in C using htslib) for manipulating next-generation sequencing data"
+LABEL website="https://github.com/samtools/samtools"
+LABEL license="https://github.com/samtools/samtools/blob/develop/LICENSE"
+LABEL maintainer="Erin Young"
+LABEL maintainer.email="eriny@utah.gov"
+
+ENV DEBIAN_FRONTEND noninteractive
```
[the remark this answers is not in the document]
kapsakcj: Additionally it will have to include an equals sign, so `ARG DEBIAN_FRONTEND=noninteractive`

Not sure why, but it didn't work without the equals sign

erinyoung:
I'll get these fixed right now. Thank you for reviewing this!

kapsakcj (approved):
Thanks for making that change. I'll merge and then run the 3 deploy workflows 👍

PR #320, opened 2022-03-23
Add image for DSK

SarahNadeau:
This PR adds DSK, a tool for k-mer counting in fasta or fastq files. It also adds test-dsk.yml and deploy-dsk.yml workflows. There are 2 commits that toggle from deploying DSK to my own docker hub to setting it up to deploy to StaPH-B's docker hub. So, here's where the image currently lives: https://hub.docker.com/repository/docker/snads/dsk. I'd love it to live in the StaPH-B docker hub moving forward, the deploy workflow is now set up for that.

  - [x] Directory structure should be name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. spades/3.12.0/Dockerfile)
- [x] Edit main README.md
- [x] Edit Program_Licenses.md
- [x] Create a simple container-specific README.md in the same directory as the Dockerfile (i.e. spades/3.12.0/README.md)
- [x] Write a GitHub actions workflow
  - [x] Should be located in .github/workflows/ and named test-<tool>.yml (i.e. .github/workflows/test-spades.yml)
 See https://github.com/SarahNadeau/docker-builds/actions/runs/2028098727.


SarahNadeau:
I am not sure why the test workflow was not run on pull request as specified [here](https://github.com/StaPH-B/docker-builds/pull/320/files#diff-a9628ef8e5ed24d5c1e5998605eaf864ba87a29b1118f1816ec7d434987060b7R6). Does the workflow show up as needing approval to a maintainer?

erinyoung:
I looked into why it's not triggering a github action.

1. It hasn't attempted the github action
2. It isn't asking for approval to start a github action

.github/workflows/test-dsk.yml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "dsk/0.0.100/Dockerfile"
+
+jobs:
+
+  build-to-test:
+    uses: StaPH-B/docker-builds/.github/workflows/build-to-test.yml@master
```
erinyoung, line 13: If you changed this to `./.github/workflows/build-to-test.yml`, does it trigger a workflow?

SarahNadeau: Indeed it does, I wonder why though. I'll take another read through the workflow docs. Anyways, should we update the template file accordingly here? https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/test-template.yml#L19

Also I did find in the docs that the workflow approval is only triggered the first time someone contributes, after that workflows run automatically.

erinyoung:
That appears to have worked. I think we need to use the relative path.

erinyoung:
In the meantime, could you add a small readme at `dsk/0.0.100/README.md`?

Generally when people get stuck on a container, they ask us how to use it instead of the container maintainer, and it makes our lives easier if there's a readme to refer them to.

erinyoung (approved):

Issue #133, opened 2020-07-13
add shigatyper


kapsakcj:
@AndrewLangvt whenever you get a chance 😃

erinyoung:
I was trying to use the [biocontainers one on quay](https://quay.io/repository/biocontainers/shigatyper?tab=tags), and it's missing samtools

kapsakcj:
Addressed in #321 , but we need to deploy the image to dockerhub & quay. Going to leave this open until that happens.

OK to trigger the deploy-shigatyper workflow?

kapsakcj:
I triggered the `deploy-shigatyper.yml` workflow manually, but it failed since quay_username and quay_robot_token were not provided. Log here: https://github.com/StaPH-B/docker-builds/runs/5796541592?check_suite_focus=true

These inputs should be optional, so I didn't expect it to fail.

We can either add those inputs to `deploy-shigatyper.yml` & re-trigger the workflow or we can somehow adjust the `build-to-deploy.yml` to truly make those inputs optional and allow the dockerhub deployment to proceed when quay_username and quay_robot_token are not provided as inputs

kapsakcj:
Closing as I've updated the `deploy-shigatyper.yml` workflow and successfully pushed the docker image to both dockerhub and quay

GH actions log here: https://github.com/StaPH-B/docker-builds/actions/runs/2100137656

PR #321, opened 2022-03-25
Adding ShigaTyper

jwarnn:
- [ x] This comment contains a description of what is in the pull request.

Adding ShigaTyper. ShigaTyper is a quick and easy tool designed to determine Shigella serotype using Illumina (single or paired-end) or Oxford Nanopore reads with low computation requirement.

- [ ] (Optional) Dockerfile is built with best practices and has been approved by a linter (such as https://hadolint.github.io/hadolint/)
- [ x] Edit main README.md
- [x ] Edit Program_Licenses.md
  - [ x] Should be located in .github/workflows/ and named test-<tool>.yml (i.e. .github/workflows/test-spades.yml)
  - [ x] Any files required for building are located in the same directory as the Dockerfile (i.e. spades/3.12.0/my_spades_tests.sh)


jwarnn:
#133 This pull request would resolve this issue.

shigatyper/2.0.1/Dockerfile
```diff
@@ -0,0 +1,108 @@
...
+RUN wget https://github.com/samtools/bcftools/releases/download/${bcftoolsVer}/bcftools-${bcftoolsVer}.tar.bz2 && \
+  tar -vxjf bcftools-${bcftoolsVer}.tar.bz2 && \
+  rm bcftools-${bcftoolsVer}.tar.bz2 && \
+  cd bcftools-${bcftoolsVer} && \
+  make && \
+  make install && \
+  cd ..
+
+# get miniconda and the shigatyper repo
+RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh && \
```
erinyoung, line 71: It looks like shigatyper is relying on conda to package it. Should we attempt to install it directly from a github download?

jwarnn: The only installation offered on the github page is a conda package. But let me see if I can use that package without using conda.

jwarnn: So I am running into some versioning problems trying to avoid conda. The repository that apt-get install uses are older versions of python, 3.5.2,  and pip. Without updating pip I get this error:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 5537: ordinal not in range(128)
And also a line that suggests an update for pip. If I then put in a line that updates pip I then get this error: 
 File "/usr/local/lib/python3.5/dist-packages/pip/_internal/cli/main.py", line 57
    sys.stderr.write(f"ERROR: {exc}")
Which I believe is caused by the pip version is for newer releases of python and no longer works with this version of python. 

I have not found a way to get a version of python that might work. It might be just as dependable to use bioconda for this tool.

erinyoung: Good to know!

PR #324, opened 2022-03-31
Add tests for Trimmomatic 0.39

SarahNadeau:
This pull request adds built-in tests for trimmomatic/0.39. 
It also fixes the path to the build-to-test.yml workflow to be local so that this workflow runs upon pull request. See https://github.com/StaPH-B/docker-builds/pull/320#discussion_r835425527. 

- [x] Ensure tool is listed in Program_Licenses.md
- [x] Ensure a simple container-specific README.md exists in the same directory as the Dockerfile (i.e. spades/3.12.0/README.md)
- [x] Update GitHub actions workflow if needed
See https://github.com/SarahNadeau/docker-builds/actions/runs/2070906094.

### DIFF
--- a/.github/workflow-templates/test-template.yml
+++ b/.github/workflow-templates/test-template.yml
@@ -16,7 +16,7 @@ jobs:
 
   # This job calls a workflow to build the image to the 'test' stage
   build-to-test:
-    uses: StaPH-B/docker-builds/.github/workflows/build-to-test.yml@master
+    uses: ./.github/workflows/build-to-test.yml
     with:
       path_to_context: "./<program name>/<program version>"  # Path to directory with Dockerfile and context, e.g. "./spades/3.12.0"
       dockerfile_name: "Dockerfile"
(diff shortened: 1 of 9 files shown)


SarahNadeau:
To whoever reviews, comments/improvements on the testing setup would be great because I'd love to use this same setup for other tools. By testing setup I mean `tests/scripts/*` to pull & run data through the tool, then python unittests to verify version and results checksums.

.github/workflow-templates/test-template.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
   # This job calls a workflow to build the image to the 'test' stage
   build-to-test:
-    uses: StaPH-B/docker-builds/.github/workflows/build-to-test.yml@master
+    uses: ./.github/workflows/build-to-test.yml
```
erinyoung, line 19: I think this is a good idea that works consistently.

erinyoung:
checksums are fancy. They might be a little overkill to require something like that for each tool in this repo.

PR #326, opened 2022-04-01
Add FLASH for merging overlapping paired-end reads

SarahNadeau:
This is the second in a series of tools we use in BSBP that we'd love to include in the StaPH-B arsenal. I kept the (maybe overkill) checksum-based tests because I can't think of an easier way to validate the output off the top of my head. Let me know if this is a problem.

  - [x] Directory structure should be name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. spades/3.12.0/Dockerfile)
- [x] Edit main README.md
- [x] Edit Program_Licenses.md
- [x] Create a simple container-specific README.md in the same directory as the Dockerfile (i.e. spades/3.12.0/README.md)
- [x] Write a GitHub actions workflow
  - [x] Should be located in .github/workflows/ and named test-<tool>.yml (i.e. .github/workflows/test-spades.yml)
 See: https://github.com/SarahNadeau/docker-builds/actions/runs/2077294504


flash/1.2.11/tests/test_controls.py
```diff
@@ -0,0 +1,23 @@
+import unittest
```
erinyoung, line 1: You're just doing checksums with this script, right?

Have you looked at using md5sum and diff? Then you wouldn't need to install python.

I think I should try this for my next container so you can see an example.

flash/1.2.11/tests/test_versions.py
```diff
@@ -0,0 +1,16 @@
...
+from subprocess import PIPE
+import re
+
+
+class TestVersions(unittest.TestCase):
+
+    def test_flash_version(self):
+        bash_cmd = "flash -v"
+        result = subprocess.run(bash_cmd, shell=True, stdout=PIPE)
+        assert re.search(b"FLASH v1.2.11\n", result.stdout)
```
erinyoung, line 12: This script is great and all... but this is how I would do it in bash/shell

```
if [ "$(flash -v)" == "FLASH v${FLASH_VER}" ] ; then echo "Correct Version" ; else echo "Version was unexpected" ; exit 1 ; fi
```

flash/1.2.11/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+WORKDIR /data
+
+FROM app as test
+ARG FLASH_VER
+
+# Get test dependencies
+RUN apt-get update && apt-get install -y \
+    bc \
+    python3 \
+    wget
```
erinyoung, line 42: why are you installing wget twice?

flash/1.2.11/tests/scripts/run_flash_pos_control.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# This script runs FLASH to merge overlapping paired-end reads.
+
+R1=/data/R1.fastq.gz
+R2=/data/R2.fastq.gz
+OUTDIR=/test_result
+mkdir -p $OUTDIR
+
+# Get read length, make minimum overlap 80% of read length
+READ_LEN=$(gunzip -c "$R1" | head -n 400 \
```
erinyoung, line 10: Since you are using test data with a known read length, is this really required?

erinyoung:
This PR is likely overly complicated and would be difficult for a APHL fellow to maintain. 

We don't really have a linting system in place, and I'm not especially picky, so I don't see any reason to reject this PR.

That being said, it'd be awesome if your future contributions could be simpler.

SarahNadeau:
I don't mind leaving this open/ submitting a new PR to fix it up per your suggestions. Like I said, I'd love to settle on a good testing format now that I could apply to lots of other tools going forward. Would you be willing to consider a fixer-upper PR with python unittests replaced by bash conditionals?

flash/1.2.11/tests/scripts/run_flash_pos_control.sh
SarahNadeau: Nope, not required. I'll remove it.

flash/1.2.11/Dockerfile
SarahNadeau: I install it twice because of the three-stage Dockerfile - the goal is to make the `app` stage as slim as possible. `builder` needs wget to get and compile the program, `app` doesn't need wget (only the program executable), and `test` does need wget again to fetch test data. This ends up saving space in the `app` stage:
Screen Shot 2022-04-04 at 9 35 02 AM

SarahNadeau: Sizes are from building the image to each stage, i.e. `docker build --target=builder -t flash-builder flash/1.2.11/`

kapsakcj:
Just ran the deploy workflow for FLASH, it's now available on dockerhub and quay

I'm not sure about the testing formats, I'm fine to leave as is as long as you are willing to maintain if it were to break in the future (likely won't break but famous last words)

Changing to a more simpler test format would be more accessible & easily understandable to other users/maintainers. Might just do this for future docker images

Issue #310, opened 2022-03-10
[Container Request]: mummer


kapsakcj:
### Contact Details

curtis.kapsak@theiagen.com

### What tool are you looking for and why?

mummer for calculating ANI values

Looks like mummer v4.x.x is a separate code repo from mummer 3.x.x and older. Probably best to go with latest version, I think?
https://github.com/mummer4/mummer

It's on [bioconda](https://bioconda.github.io/recipes/mummer4/README.html?highlight=mummer#package-package%20&#x27;mummer4&#x27;) and [biocontainers exist for mummer4 on quay.io](https://quay.io/repository/biocontainers/mummer4?tab=tags), but would be nice to have a StaPH-B docker image available too.

kapsakcj:
#327  PR should close this Issue, but I'd like to keep it open for now. I'll create a second dockerfile that includes the RGDv3 fasta files to be used for ANI. I think this will be very useful for those in the community that are working with enteric pathogens

kapsakcj:
started dev branch `cjk-mummer-RGD` for this. Needs testing & polishing before starting a PR

Additionally started PR #399 to add RGDv2 to the fastANI docker image for similar functionality

kapsakcj:
closed via #405

PR #327, opened 2022-04-01
MUMMer Tool 

jwarnn:
- [X ] This comment contains a description of what is in the pull request.

This pull request is to resolve this issue: #310 
This tool has alot of functionality and is easy to use. I would just like to note that the command mummerplot must use the option --terminal png unless you are working within the container interactively. I hope this tool is helpful!

- [ ] (Optional) Dockerfile is built with best practices and has been approved by a linter (such as https://hadolint.github.io/hadolint/)
- [ x] Edit main README.md
- [ x] Edit Program_Licenses.md
  - [ x] Should be located in .github/workflows/ and named test-<tool>.yml (i.e. .github/workflows/test-spades.yml)
  - [ x] Any files required for building are located in the same directory as the Dockerfile (i.e. spades/3.12.0/my_spades_tests.sh)


mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+  wget https://mummer4.github.io/tutorial/exampleFiles/2.1/in/H_pyloriJ99_Eslice.fasta && \
+  wget https://mummer4.github.io/tutorial/exampleFiles/2.2/in/B_anthracis_Mslice.fasta && \
+  wget https://mummer4.github.io/tutorial/exampleFiles/2.2/in/B_anthracis_contigs.fasta && \
+  wget http://mummer.sourceforge.net/examples/data/H_pylori26695_Eslice.fasta && \
+  wget http://mummer.sourceforge.net/examples/data/H_pyloriJ99_Eslice.fasta
+RUN mummer -mum -b -c H_pylori26695_Eslice.fasta H_pyloriJ99_Eslice.fasta > mummer.mums
+RUN mummerplot -x "[0,275287]" -y "[0,265111]" --terminal png -postscript -p mummer mummer.mums
+RUN nucmer  -c 100 -p nucmer B_anthracis_Mslice.fasta B_anthracis_contigs.fasta
+RUN show-snps -C nucmer.delta > nucmer.snps
+RUN promer -p promer_100 -c 100  H_pylori26695_Eslice.fasta H_pyloriJ99_Eslice.fasta
```
kapsakcj, line 74: Just a suggestion - to reduce the number of layers in the `test` stage, you could combine lines 62-74 into one `RUN` block. 

Would also suggest adding `nucmer -v` and `promer -v` after line 63 if those flags exist (I'm not familiar with the tools)

kapsakcj:
Thank you so much for the PR! Look forward to reviewing when I have some free time

mummer/4.0.0/Dockerfile
jwarnn: I grouped the test layers into one block. numer and promer do not hace the -v flags.

mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,75 @@
+# FROM defines the base docker image. This command has to come first in the file
+# The 'as' keyword lets you name the following stage. We use `app` for the production image
+FROM ubuntu:xenial as app
```
kapsakcj, line 3: I would recommend upgrading to `ubuntu:focal` since `ubuntu:xenial` is EOL

mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+
+# ARG sets environment variables during the build stage
+ARG MUMMER_VER="4.0.0"
+
+# LABEL instructions tag the image with metadata that might be important to the user
+LABEL base.image="ubuntu:xenial"
```
kapsakcj, line 9: if you upgrade the base image, please update this LABEL to reflect the new base image

mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+LABEL dockerfile.version="1"
+LABEL software="mummer"
+LABEL software.version=$MUMMER_VER
+LABEL description="MUMmer is a versatile alignment tool for DNA and protein sequences."
+LABEL website="https://github.com/StaPH-B/docker-builds"
+LABEL license="https://github.com/StaPH-B/docker-builds/blob/master/LICENSE"
```
kapsakcj, line 15: lines 14 and 15 should list the website and license of the mummer software

mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+ gcc \
+ make \
+ perl \
+ bzip2 \
+ gnuplot \
+ ca-certificates \
+ gawk \
+ curl \
+ sed \
+ sudo \
```
kapsakcj, line 39: I would recommend removing `sudo` here. It is not necessary since when the docker image is built, it is by default built as the `root` user. 

Additionally you can remove `sudo` from line 51

mummer/4.0.0/Dockerfile
```diff
@@ -0,0 +1,75 @@
...
+  rm mummer-${MUMMER_VER}rc1.tar.gz && \
+  cd mummer-${MUMMER_VER}rc1 && \
+  ./configure && \
+  make && \
+  make install && \
+  sudo ldconfig && \
+  cd .. && \
+  mkdir /data
+
+ENV PATH="/usr/local/bin:$PATH" \
```
kapsakcj, line 55: line 55 may be unnecessary, I'm 99% sure `/usr/local/bin` is in PATH be default. Doesn't hurt to keep it though, just sets the priority for that directories over all others in PATH

kapsakcj:
Hi John - thanks for the PR. I've made a couple of suggestions for improvement - let me know if there is anything I can clarify or if you have any questions.

both GH actions workflows look great. We can already see that the `test-mummer.yml` works as expected.

I've already created the repo on quay, so when we merge the PR and one of the maintainers triggers the `deploy-mummer.yml` workflow, the docker image should automatically be pushed to both dockerhub and quay 🤞

The dockerhub repo should automatically be created upon deployment, but we can create it manually if that is not the case.

mummer/4.0.0/Dockerfile
jwarnn: I did have a question about this. In Dockerhub the ubuntu image does have a latest tag; do we avoid using this for our Staph-b docker-builds because as ubuntu gets updated it could cause some incompatibilities ?

kapsakcj: Yes, `ubuntu:latest` will change a lot more frequently than `ubuntu:focal` will. Changes to `ubuntu:focal` will be restricted to that particular release of Ubuntu and will mostly include bug fixes and security patches (which we would welcome 👍 )

kapsakcj:
Awesome! We appreciate the contributions.

Not sure why, but the mummer test has been running for over 2hrs now, so I'm going to kill the job and restart. Unfortunately it's not showing any logging information so it's hard to tell what is going on. Hopefully a re-start will fix it

kapsakcj:
Ah, I know exactly what happened. `apt` is essentially asking for a user to enter a time zone. We need to make it run without user input.

Could you please add this line of code right before the `apt-get update && apt-get install [...]` steps? Around line 18-19 of the Dockerfile

```Dockerfile
# to prevent tzdata from asking for a region during apt updates; ARG so that variable only 
# persists at buildtime
ARG DEBIAN_FRONTEND=noninteractive
```

kapsakcj:
et voila! GH actions workflow ran successfully 🎆 

@erinyoung do you want to give this a quick review or are we OK to merge and deploy?

kapsakcj (approved):

PR #347, opened 2022-04-20
fix badges on README by switching to badgen.net 

### LINKED ISSUE #346: [Bug]: Fix badges for # docker pulls
### Contact Details

slack, email, snail mail

### What container were you trying to use, and how were you attempting to use it?

Lots if not all of the badges from shields.io on the main README displaying the number of docker pulls are now showing "Invalid" instead of the number of pulls.

I spent more time then I care to admit when troubleshooting on my fork `kapsakcj/docker-builds` but I could never get all of them to display the number of pulls from dockerhub appropriately.

Even when I tried to test here: https://shields.io/category/downloads with various staphb dockerhub repos, I would still get the "Invalid" response for some instead of the expected # of pulls. I had to play around with the URLs and even after a few hours of attempts, I couldn't get them all fixed in a stable manner.

I think it may be related to this issue on the Shields repo: https://github.com/badges/shields/issues/7583

But I don't know enough about the Dockerhub API and how to fix these appropriately.

This is a tough one, in my opinion. Not sure what to do here. Feedback and suggestions are very welcome.

### Relevant log output

_No response_

kapsakcj:
This PR should fix #346 .

Currently on the staph-b/docker-builds `master` branch, many if not all badges for docker pulls show `Invalid` instead of the expected number of pulls.

I don't want to dig too much further with Shields.io & try to troubleshoot so I went the somewhat brute force route of switching to a different provider/server of badges.

The new provider is https://badgen.net/ and you can find more documentation there.

This PR replaces all sheilds.io links with badgen.net links and you can see the final, rendered product on my fork here (`master` branch): https://github.com/kapsakcj/docker-builds


erinyoung (approved):
I checked it out on https://github.com/kapsakcj/docker-builds, and it looks really nice.

kapsakcj:
Definitely squashing the commits here. please don't look at my mess 🙈

Issue #365, opened 2022-05-10
[Repo Request]: update manual_deploy.yml


kapsakcj:
### Contact Details

curtis.kapsak@theiagen.com

### What functionality is this repo missing?

I just merged #355 but afterwards realized that I'd like something adjusted. I want to separate the `version` of the tool used to locate the dockerfile and the `docker image tag` used for deployment to dockerhub/quay

Currently I can specify the `tool` and `version` of the tool that I want to deploy, and that information is used to find the dockerfile `/tool/version/Dockerfile` but I can't specify the `docker image tag`.

So for example, I'm forced to use this docker image name and tag `staphb/kraken2:2.1.2` (dockerfile located at `/kraken2/2.1.2/Dockerfile`) and I can't specify a custom image tag, like `staphb/kraken2:2.1.2-no-db` . Good for enforcing the style we prefer, but not good for when we want a custom tag like `staphb/pangolin:4.0.6-pdata-1.8` The docker image tag does not match the explicit version of Pangolin.

CC @SarahNadeau for awareness. I should have done a more thorough review before merging your PR, but this should be relatively straight forward to add. I will make a PR when I'm able to

kapsakcj:
Working on this over on my fork https://github.com/kapsakcj/docker-builds/

Will also add an additional feature to where we can specify if we want to overwrite the `:latest` default docker image tag or not. Planning to make it a boolean like the push_quay input.

Usually we want to do that, but I can see situations where we re-build older versions and don't want to overwrite the `:latest` tag with an older version.

kapsakcj:
Addressed via #366

PR #366, opened 2022-05-11
manual_deploy.yml and build-to-deploy.yml enhancements

kapsakcj:
- [ ] This comment contains a description of what is in the pull request.

This PR is in response to #365 . It updates 2 files:

### 1. `.github/workflows/manual_deploy.yml`

- Updates user prompts to be more informative/descriptive
- added prompt (checkbox) for choosing to overwrite default docker image tag `repository/<tool>:latest` or not
- added prompt for user to define custom `docker_image_tag` to use instead of `version` ➡️ `repository/tool:<docker-image-tag>`
- updated jobs with new user input `docker_image_tag` and boolean for overwriting latest docker image tag

### 2. `.github/workflows/build-to-deploy.yml`

- added new boolean input `push_latest_tag`
- added new job in workflow for extracting metadata while building an image and tagging multiple times. This was done to prevent errors while pushing to quay. More info here: https://github.com/docker/build-push-action/issues/452 and the failures I were seeing: https://github.com/kapsakcj/docker-builds/actions/runs/2304558965
`buildx failed with: error: failed to solve: failed commit on ref "manifest-sha256:30317537135a5a50fb6b4b3dc4300b81facb1374be464bac541cd21bc2b1576d": cannot reuse body, request must be retried`
- added new jobs in workflow for pushing to dockerhub and quay for `:latest` tag. Only runs if user checks box to overwrite latest tag
- adjusted previous dockerhub and quay build jobs to only tag and push images with user-defined tag (since `:latest` tags are separate jobs now)

I'll follow up with screenshots and links to dockerhub and quay showing the functionality with our trusty test workhorse `spades` 3.15.4. Having the ability to set a custom docker image tag made it really easy to test.

- [ ] Update relevant GitHub actions workflow files
- [ ] Any files required for building are located in the same directory as the Dockerfile (i.e. spades/3.12.0/my_spades_tests.sh)
- [ ] Have successfully run the workflow "Test <program name> image" in your forked repository


kapsakcj:
I tried 4 combinations with the `spades` 3.15.4 dockerfile with this updated `manual_deploy.yml` GH actions workflow:

1. Only user-defined tag to dockerhub. https://github.com/kapsakcj/docker-builds/runs/6381653030?check_suite_focus=true
spades-onlydhub-without-latest

2. user-defined and latest tags, only dockerhub https://github.com/kapsakcj/docker-builds/runs/6381672591?check_suite_focus=true
spades-onlydhub-with-latest

3. Only user-defined tag, both dockerhub and quay https://github.com/kapsakcj/docker-builds/runs/6381692333?check_suite_focus=true
spades-dhub-and-quay-without-latest

4. Both user-defined and latest tags, both dockerhub and quay https://github.com/kapsakcj/docker-builds/runs/6381795904?check_suite_focus=true
spades-dhub-and-quay-with-latest

kapsakcj:
We can see the custom tags on my kapsakcj/spades repo on quay. The `latest` tag matches the same image as `3.15.4-dhub-and-quay-overwrite-latest` tag (same hash and modification time) https://quay.io/repository/kapsakcj/spades?tab=tags

Same goes for dockerhub: https://hub.docker.com/r/kapsakcj/spades/tags

kapsakcj:
@SarahNadeau I can't make you a reviewer (I think you need to be part of the StaPH-B github org?), but would appreciate any feedback you might have!

erinyoung:
I have two questions:
- Is there a default value for repository already there? I think it's great that it can be specified for other repos, but it'd be nice to have staphb already there.
- Is there a way to autopopulate the tag to be the same as the version? I think it'd be better if there were a flag of something like do-not-use-version for tag, and then specify what the tag would be.

kapsakcj:
1. Yes, `staphb` is the default repository, but can be changed if folks are running on their fork and want to push to their personal dockerhub/quay repos

https://github.com/kapsakcj/docker-builds/blob/2ff05c2e0c355e2b6e2e8e936e46d71da6f557ad/.github/workflows/manual_deploy.yml#L29

2. I'm not sure, I'll look into it and see what's possible

kapsakcj:
I can't figure out how to populate user input from the version field into the docker image tag field. 

Hopefully the description is worded strongly enough to encourage the maintainers (us, lol) to stick to using the version as the docker image tag, unless truly necessary.

Here's the documentation on workflow_dispatch inputs in case you'd like to read further. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs

Here's what it a blank input form looks like, let me know if you want to change/add anything
image

SarahNadeau:
Thanks for fixing this up @kapsakcj! I had 2 ideas for a cleaner input form that avoids typos in entering the version twice, but what you have already looks quite sensible to me if you want to just use that.

Idea 1: use a conditional job so that if tag input = empty string, then use the version as the tag. Then another conditional job so that if tag input != empty string, then use the tag input as tag. Haven't thought this all the way through but seems possible.

Idea 2: you can still manually deploy images, right? I.e., what you did before this repo had any GitHub actions workflows? You could say special cases where the tag differs from the version must be done manually. Then you just have to handle the typical case where version=tag in the github actions workflow.

kapsakcj:
thanks for the feedback, Sarah! I like idea 1. I'll work on adding a conditional job. That sounds less error prone and better sets the default image tag as the version

idea 2 - yes we can manually deploy, but I'd rather have the ability to deploy via GH actions with custom image tags. It's a bit cumbersome to use dockerhub or other methods. It's just too easy with GH actions and I'm hooked 😆

PR #375, opened 2022-05-19
Plasmidfinder 2.1

jwarnn:
- [ x] This comment contains a description of what is in the pull request.

- [ x] (Optional) Dockerfile is built with best practices and has been approved by a linter (such as https://hadolint.github.io/hadolint/)
- [ x] Edit main README.md
- [ x] Edit Program_Licenses.md
  - [x ] Should be located in .github/workflows/ and named test-<tool>.yml (i.e. .github/workflows/test-spades.yml)
  - [x ] Any files required for building are located in the same directory as the Dockerfile (i.e. spades/3.12.0/my_spades_tests.sh)


.github/workflows/deploy-plasmidfinder.yml
```diff
@@ -0,0 +1,43 @@
+# This caller workflow tests, builds, and pushes the image to Docker Hub.
```
erinyoung, line 1: This file isn't needed anymore. 

Yes, we need to update the documentation.

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+ARG PLASMIDFINDER_VER="2.1"
+
+# LABEL instructions tag the image with metadata that might be important to the user
+# Optional, but highly recommended
+LABEL base.image="debian:stretch"
+LABEL dockerfile.version="1"
+LABEL software="plasmidfinder"
+LABEL software.version=$PLASMIDFINDER_VER
+LABEL description="Identifies plasmids in total or partial sequenced isolates of bacteria."
+LABEL website="https://github.com/StaPH-B/docker-builds"
```
erinyoung, line 16: The website and license need to be updated to plasmidfinder instead of the docker-builds repo

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+LABEL license="https://github.com/StaPH-B/docker-builds/blob/master/LICENSE"
+LABEL maintainer="John Arnn"
+LABEL maintainer.email="jarnn@utah.gov"
+
+
+ENV DEBIAN_FRONTEND noninteractive,Teletype
```
erinyoung, line 22: It's better to use
```
ARG DEBIAN_FRONTEND="noninteractive,Teletype"
```
so that it disappears after the build

erinyoung: Also, what is the 'Teletype' for?

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+# RUN executes code during the build
+# Install dependencies via apt-get or yum if using a centos or fedora base
+RUN set -xe && \
```
erinyoung, line 26: what is the 
```
set -xe
```
for?

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    apt-get update -y && apt-get install -y \
+    git \
+    apt-utils \
+    wget \
+    python3-pip \
+    ncbi-blast+ \
+    gzip \
+    libz-dev
+    #rm -rf /var/cache/apt/* /var/lib/apt/lists/*
```
erinyoung, line 35: Could you add 
```
rm -rf /var/lib/apt/lists/* && apt-get autoclean
```
to the end? This removes unused packages

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+# Install python dependencies
+RUN pip3 install --upgrade pip && \
+    python3 -m pip install tabulate && \
+    pip3 install -U biopython==1.73 cgecore
+
+# Install kma
+RUN git clone --branch 1.0.1 --depth 1 https://bitbucket.org/genomicepidemiology/kma.git
```
erinyoung, line 43: Replacing this with 
```
ARG kma_ver=1.0.1
wget https://bitbucket.org/genomicepidemiology/kma/get/{kma_ver}.tar.gz && etc
```
would be easier to update in the future if a new kma version is needed.

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+RUN cd kma && make && \
+    mv kma* /bin/
+
+RUN git clone https://bitbucket.org/genomicepidemiology/plasmidfinder.git && \
```
erinyoung, line 47: Instead of git clone, something like
```
wget https://bitbucket.org/genomicepidemiology/plasmidfinder/get/${PLASMIDFINDER_VER}.tar.gz
```
is better because it specifies the version

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    cp plasmidfinder/plasmidfinder.py /usr/src/plasmidfinder.py && \
```
erinyoung, line 48: Instead of copying this file to somewhere new, you should be able to just add this directory to path
```
ENV "PATH:/plasmidfinder"
```
Or something similar

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    mv plasmidfinder_db database
+
+ENV PATH $PATH:/usr/src
+
+WORKDIR /data
+
+FROM app as test
+
+RUN wget ftp://ftp.sra.ebi.ac.uk/vol1/fastq/ERR193/000/ERR1937840/ERR1937840.fastq.gz
+RUN gzip -d ERR1937840.fastq.gz && chmod 755 ERR1937840.fastq && mkdir test && \
```
erinyoung, line 69: if you need to chmod the test files, they won't work for the general user

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+RUN git clone https://bitbucket.org/genomicepidemiology/plasmidfinder.git && \
+    cp plasmidfinder/plasmidfinder.py /usr/src/plasmidfinder.py && \
+    mv plasmidfinder/test/database/ /database && \
+    mv plasmidfinder/test* /test/
+
+RUN chmod 755 /usr/src/plasmidfinder.py && \
+    chmod 755 test/test.sh
+
+RUN git clone https://bitbucket.org/genomicepidemiology/plasmidfinder_db.git && \
```
erinyoung, line 55: This also needs a version

plasmidfinder/1.0/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+    cd plasmidfinder_db && \
+    python3 INSTALL.py kma_index && \
+    cd .. && \
+    rm -rf database && \
+    mv plasmidfinder_db database
```
erinyoung, line 60: Instead of moving this directory, setting up 

```
ENV PLASMID_DB=<path to db>
```
would be better

erinyoung (changes requested):
This looks like this is off to a great start! 

It looks like you started with the dockerfile in the plasmidfinder gitbucket repo (which is a great place to start). I think you need to be more specific about your versions that get downloaded (and if you download everything with wget, you might not need to install git).

plasmidfinder/1.0/Dockerfile
jwarnn: plasmidfinder.py uses the flag -p as the path to database and it has the default '/database'.  The idea was the mv command renames the database downloaded from the git pull so that the user has a workable default if they do not provide a database of their own to the script.

plasmidfinder/1.0/Dockerfile
jwarnn: Updated to correct websites.

plasmidfinder/1.0/Dockerfile
jwarnn: I was using the dockerfile on the tools website as a template. It dose not seem to do anything and I couldn't find anything about it with a google search. So I took it out.

plasmidfinder/1.0/Dockerfile
jwarnn: Again from the tools website dockerfile. Set will display shell variables and commands. This is unneeded and removed.

plasmidfinder/1.0/Dockerfile
jwarnn: Removed commented line and added suggested line.

plasmidfinder/1.0/Dockerfile
jwarnn: Implemented this suggestion.

plasmidfinder/1.0/Dockerfile
jwarnn: Implemented this suggestion.

plasmidfinder/1.0/Dockerfile
jwarnn: Implemented this suggestion.

plasmidfinder/1.0/Dockerfile
jwarnn: This chmod 755 is unneeded as the test works without it. Removed.

plasmidfinder/1.0/Dockerfile
jwarnn: Implemented this suggestion.

erinyoung (approved):

erinyoung (changes requested):
I KNOW that I just approved it. I downloaded the dockerfile and it worked for *me*.

Then I realized that the test didn't kick off. It looks like there's a small typo in the directory path in 'test-plasmidfinder.yml'

jwarnn:
I updated the test-plasmidfinder.yml and it should work now.

erinyoung (approved):

PR #378, opened 2022-06-01
Adding faidx preemptively to the container

erinyoung:
Sometimes shigatyper still runs into permissions errors with singularity because the reference fasta file can't be indexed.


kapsakcj (approved):
Looks good, will deploy to dockerhub and quay, overwriting the existing docker image tag

kapsakcj:
@erinyoung I'll let you delete the branch `erin-shigatyper` if you are done with it

PR #476, opened 2022-10-24
ncbi-amrfinderplus image path

arodzh-sudo:
Small fix of the ncbi-amrfinderplus image path


kapsakcj (approved):
looks good, thanks for the fix 👍

PR #577, opened 2023-02-02
mafft version 7.505

erinyoung:
Pull Request (PR) checklist:
- [X] Include a description of what is in this pull request in this message.
- [X] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/samtools/1.15` )
- [X] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. `spades/3.12.0/Dockerfile`)
   - [X] (optional) All test files are located in same directory as the Dockerfile (i.e. `shigatyper/2.0.1/test.sh`)
- [X] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `spades/3.12.0/README.md`)
   - [X] If this README is longer than 30 lines, there is an explanation as to why more detail was needed
- [X] Dockerfile includes the recommended [LABELS](https://github.com/StaPH-B/docker-builds/blob/master/dockerfile-template/Dockerfile#L8-L18) 
- [X] Main [README.md](https://github.com/StaPH-B/docker-builds/blob/master/README.md) has been updated to include the tool and/or version of the dockerfile(s) in this PR
- [X] [Program_Licenses.md](https://github.com/StaPH-B/docker-builds/blob/master/Program_Licenses.md) contains the tool(s) used in this PR and has been updated for any missing

According to https://mafft.cbrc.jp/alignment/software/
> Versions 7.463 – 7.486 had a serious bug in the FFT-NS-i option; unnecessarily much memory was requested and the calculation sometimes failed.  This bug also affected the --auto option.  Please use 7.487 or higher.  (2021/Jul/25)

This has been on my to-do list, and I've finally updated the mafft version

PR #578, opened 2023-02-02
Update Ivar to 1.3.2

erinyoung:
Pull Request (PR) checklist:

There's a new version of iVar out. According to https://github.com/andersen-lab/ivar/releases/tag/v1.3.2, 
> New parameter added to consensus calling to allow insertions to be called at a different threshold than consensus.
a. -c Minimum insertion frequency threshold(0 - 1) to call consensus. (Default: 0.8)
> Change primer sort to binary search in primer trimming code.
> In trimming code, keep deletions adjacent to soft-clipped bases from the primer.
> Adds in gene to the GFF_FEATURE column of output for ivar variants if GFF file is provided as input.
> Update handling of strings for compatibility with newer versions of GCC
> In ivar trim, if a primer pair file is provided as input, use that information to trim primer sequences.

So.. there's a new parameter for calling consensus for insertions, and it's compatible with newer installations of GCC. As such here are the things I added to the dockerfile

1. Simplified path to htslib and samtools
2. added `--no-install-recommends` to apt-get layer
3. installs autotools-dev, automake, ca-certificates, and procps with apt-get
4. moved 'ENV' for aesthetic reasons
5. added test layer - for now

I am a little worried about the test layers because ivar trim isn't the faster process, and I've never gotten samtools mpileup to work for with github actions (requires 8 G RAM), but it works fine on my machine... so I've reduce the samtools mpileup depth to 40 and I'm hoping for the best.


erinyoung:
Ah... perfect... I forgot to switch branches. Review this one after [the ](https://github.com/StaPH-B/docker-builds/pull/577)

erinyoung:
Nevermind!!!! It looks like this version is no longer available : https://github.com/andersen-lab/ivar

erinyoung:
And... it's available again. Or has been for an hour.

ivar/1.3.2/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+  git \
+  bedtools \
+  python3-pip \
+  vim \
+  nano && \
+  apt-get autoclean && rm -rf /var/lib/apt/lists/*
+
+# installing htslib
+RUN wget https://github.com/samtools/htslib/releases/download/${HTSLIBVER}/htslib-${HTSLIBVER}.tar.bz2 && \
+  tar xvf htslib-${HTSLIBVER}.tar.bz2 && \
```
kapsakcj, line 48: I thought you needed `tar -j` for bzip2-compressed files? looks like in uncompressed fine without it 🤷 

Also, the `tar` command is quite verbose - might be good to remove the `v` option from here (and other lines) to shorten the build output

kapsakcj:
I'm fine to leave the test as is, `ivar trim` seemed to be the slowest step as it took about 5.6min out of total 10min to run the GHActions workflow. We could decrease even further by using even smaller input FASTQs.

Fine by me but I'm OK to merge as the PR stands now.

I made a couple of small changes to the ivar-specific README.md and updated my contact email on the dockerfile

erinyoung:
Thanks!

I removed the v from the tar command when extracting ivar.

It doesn't look like there's other changes I need to make. 

I look forward to you merging this and deploying to dockerhub!

kapsakcj:
deploy workflow launched: https://github.com/StaPH-B/docker-builds/actions/runs/4187717558

@erinyoung I'll let you delete the dev branch if you're done with it

PR #627, opened 2023-02-27
adding ivar 1.4

erinyoung:
Supposedly the lengthy time of ivar trim has been resolved (https://github.com/andersen-lab/ivar/releases/tag/v1.4).

EDIT: the following statement does not, yet, apply, but could in a future release.

<del>
Also of note, 
> Alignment file (Sorted/Unsorted SAM/BAM) can now be piped into ivar trim. Further, if no prefix using -p is specified, ivar trim will output trimmed SAM to stdout. This enables going from alignment to consensus using a one-liner. For example,
bwa mem idx read1.fq reads2.fq | ivar trim -b test.bed -x 3 -m 30 | samtools sort - | samtools mpileup -aa -A -Q 0 -d 0 - | ivar consensus -p test_consensus -m 10 -n N -t 0.5

So I'm really tempted to include bwa and minimap2 in the ivar container.

I have left them out for now.

</del>

Pull Request (PR) checklist:
- [X] Include a description of what is in this pull request in this message.
- [X] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/samtools/1.15` )
- [X] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. `spades/3.12.0/Dockerfile`)
   - [X] (optional) All test files are located in same directory as the Dockerfile (i.e. `shigatyper/2.0.1/test.sh`)
- [X] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `spades/3.12.0/README.md`)
   - [X] If this README is longer than 30 lines, there is an explanation as to why more detail was needed
- [X] Dockerfile includes the recommended [LABELS](https://github.com/StaPH-B/docker-builds/blob/master/dockerfile-template/Dockerfile#L8-L18) 
- [X] Main [README.md](https://github.com/StaPH-B/docker-builds/blob/master/README.md) has been updated to include the tool and/or version of the dockerfile(s) in this PR
- [X] [Program_Licenses.md](https://github.com/StaPH-B/docker-builds/blob/master/Program_Licenses.md) contains the tool(s) used in this PR and has been updated for any missing

PR #640, opened 2023-03-21
adding ivar 1.4.1 & 1.4.2

erinyoung:
iVar has some updates, including a big fix for 1.4. I have therefore closed https://github.com/StaPH-B/docker-builds/pull/627 because this would be the image that I'd be using.

Supposedly the lengthy time of ivar trim has been resolved (https://github.com/andersen-lab/ivar/releases/tag/v1.4), and the outputs from an alignment file can be piped directly into iVar trim. I have tested this locally, but this takes a long time without allocating several CPUs to alignement.

From 1.4's release notes:
> Alignment file (Sorted/Unsorted SAM/BAM) can now be piped into ivar trim. Further, if no prefix using -p is specified, ivar trim will output trimmed SAM to stdout. This enables going from alignment to consensus using a one-liner. For example,
bwa mem idx read1.fq reads2.fq | ivar trim -b test.bed -x 3 -m 30 | samtools sort - | samtools mpileup -aa -A -Q 0 -d 0 - | ivar consensus -p test_consensus -m 10 -n N -t 0.5

So I have included minimap2 and bwa in the container. They are both installed with apt-get, which seems to be stable.

I have also included a use case on how to use piping into and out of ivar trim. The example is with bwa, but should work with minimap2 as well. I wanted to include this use-case in the test stage, but since it took over 45 minutes on my computer, it's likely going to take too long for GA.

Pull Request (PR) checklist:


kapsakcj:
Happy to review, but wanted to note that ivar 1.4.2 was released recently for a bugfix with `ivar trim -f`. Any chance that you use that option and want to include that bugfix? 

https://github.com/andersen-lab/ivar/releases/tag/v1.4.2

I'm wondering we should simply update the dockerfile in this PR to use 1.4.2 instead of 1.4.1

thoughts?

erinyoung:
My thoughts? 

This PR now contains both 1.4.1 and 1.4.2. 

We can push both to dockerhub and quay, but use 1.4.2 as the latest one.

Right now 1.4.2 has a test in which fastq files are aligned with bwa and piped into ivar, which takes *forever*.

If this takes too long, I really should subsample the fastq files until they're a reasonable-enough size for GA.

kapsakcj:
Sounds good, thanks for adding the additional dockerfile for 1.4.2 👍 

Suggested changes:

For `ivar/1.4.1/Dockerfile`:

- bedtools installed via `apt` for [ubuntu:jammy is now updated to v2.30.0](https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=bedtools&searchon=names), so line 24 needs to be updated. 
  - line 11 on `ivar/1.4.1/README.md` should also reflect this change
  - the same ^ applies for `ivar/1.4.2/Dockerfile` and `ivar/1.4.2/README.md`
- I recommend that we remove `vim` and `nano` from `apt-get install` step in both dockerfiles. I believe they were carried over from Karthik's original Dockerfile and are not necessary. Deleting those from the `apt-get install` step resulted in 43 MB savings in uncompressed docker image size:

```bash
$ docker images | grep ivar
erin/ivar  1.4.2-noNanoVim   6f350a81cd0b   About a minute ago   751MB
erin/ivar  1.4.2             28809e6cb839   14 minutes ago       794MB

```

Regarding the new test (big oneliner) - I agree, I uncommented the test & it took forever to run for v1.4.1, so I ended up cancelling before it even finished. 

BUT it seems that with 1.4.2, the test finishes relatively quickly? The one-liner test is uncommented in the dockerfile (not sure if you did that on purpose or not), but it ran in a reasonable amount of time - see time to complete GHActions workflows. I'm OK to keep it that way for 1.4.2

Thanks for adding `minimap2` and `bwa` to the docker image! This is one example where the benefits outweigh the cons of including dependencies that are not absolutely required, but highly recommended to save overhead costs for modular workflows.

kapsakcj (changes requested):
see previous comment for requested changes

erinyoung:
I think I have made the requested changes.

I have
- changed the readme and comments to indicate bedtools is actually version 2.30.0 (thank you for catching that!)
- removed nano and vim on both 1.4.1 and 1.4.2
- removed bwa and minimap2 in 1.4.1 and corresponding text in readme and comments

erinyoung:
The tests failed, so I made additional changes. 

Cecret's directory structure changed, so the reference file could not be downloaded for the test stage. The reference fasta is now downloaded from artic.

kapsakcj:
I'll let you delete the dev branch if you are finished with it.

deploy workflows have been queued: https://github.com/StaPH-B/docker-builds/actions/runs/4800386170

and 

https://github.com/StaPH-B/docker-builds/actions/runs/4800388104

PR #649, opened 2023-04-06
update contributor list

kapsakcj:
just fixing the contributor list on the main `/README.md`. No other changes

fixed a broken link. added Ines since I forgot to add her on the last PR


erinyoung:
Is this everyone now?

erinyoung (approved):

kapsakcj:
> Is this everyone now?

I hope so, but we can always do another PR if there are other's we accidentally left out

erinyoung:
I think there should be 55+ people listed, so I'm sure it's still missing some, and it might be helpful to alphabetize it at some point.

But I think this is a step to keep it current-as-possible.

PR #681, opened 2023-06-14
HMMER version update 3.3.2 

golden75:
Pull Request (PR) checklist:
- [ X] Include a description of what is in this pull request in this message.
- [ X] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/samtools/1.15` )
- [ X] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number (i.e. `spades/3.12.0/Dockerfile`)
   - [ ] (optional) All test files are located in same directory as the Dockerfile (i.e. `shigatyper/2.0.1/test.sh`)
- [ X] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `spades/3.12.0/README.md`)
- [ X] Dockerfile includes the recommended [LABELS](https://github.com/StaPH-B/docker-builds/blob/master/dockerfile-template/Dockerfile#L8-L18) 
- [X ] Main [README.md](https://github.com/StaPH-B/docker-builds/blob/master/README.md) has been updated to include the tool and/or version of the dockerfile(s) in this PR
- [ X] [Program_Licenses.md](https://github.com/StaPH-B/docker-builds/blob/master/Program_Licenses.md) contains the tool(s) used in this PR and has been updated for any missing


golden75:
update HMMER version 3.3.2

kapsakcj (changes requested):
Hi Neranjan, thanks for the PR. Nice work!

I have a few requests for changes, please make the changes and push the commit(s) when you get the chance:

`README.md` changes
- [x] (optional) Please add yourself to the list of authors/maintainers at the bottom of the `README.md` if you would like to list a link to your github profile. Not required, but we'd like to give you credit for updating this dockerfile!
- [x] Please add a new bullet for the newer version of hmmer 3.3.2 and keep the existing bullet for hmmer 3.3. We want to list both versions since there will be 2 docker image tags available after the PR is merged & new docker image is deployed. Additionally, we can add a relative links to the subdirectories so it is easy for users to find the hmmer-specific documentation. Please update the code to look like this:
```
<ul><li>[3.3](hmmer/3.3/)</li><li>[3.3.2](hmmer/3.3.2/)</li></ul>
```

`hmmer/3.3.2/Dockerfile` changes
- [x] Line 16, please change `ENV` to `ARG`. We want this variable to be only available during image build time, and not permanently set in the final docker image. Users should use the hmmer `--version` flags to check the version instead of relying on an arbitrary environmental variable.
- [x] Line 35, please change the line so that the default PATH variable is higher priority than `/hmmer-${HMMERVERSION}/src`. We want users to use the executable hmmer commands in `/usr/local/bin` instead since those are the final executables created by the `make install` command. Line 39 should be changed to this:
```Dockerfile
ENV PATH="${PATH}:/hmmer-${HMMERVERSION}/src" \
```
- [x] Combine the commands in lines 49 and 51 into a single `RUN` instruction by using the `&&` operator. 

FYI - the automated tests via GitHub actions have passed already ✅ and we should see them pass again after updating the dockerfile

kapsakcj (approved):
thank you for making those changes!

I'm going to merge the PR & kick of the docker image deploy workflow

kapsakcj:
The docker image deployment workflow can be viewed here: https://github.com/StaPH-B/docker-builds/actions/runs/5271210377

We should see new docker image tags appear on dockerhub and quay shortly:

https://hub.docker.com/r/staphb/hmmer/tags

https://quay.io/repository/staphb/hmmer?tab=tags

Thank you for the PR!

PR #692, opened 2023-07-18
Create CODE_OF_CONDUCT.md

erinyoung:
Codes of conduct seem really popular these days, so I think we should have one, too. 

I haven't noticed any bad actors, but that doesn't mean that they haven't existed in the StaPH-B community or won't exist in the future.

This PR is just adding a code of conduct to this repo. I used a template, so I won't take it personally if others feel like things need to be re-worded.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,94 @@
+# Citizen Code of Conduct
+
+## 1. Purpose
+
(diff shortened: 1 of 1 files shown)


kapsakcj:
This looks great, I'm just wondering who should be responsible for handling infractions of the code of conduct? In other words, who should someone contact?

StaPH-B steering committee members? Or the admins of this github repo? both?

I feel like the steering committee may be more appropriate, but they not be privy to all the interactions of people contributing to this repo.

I'm happy to be listed as someone to contact but I don't want to add other people to the contact list without their knowledge and approval.

Thoughts? @StaPH-B/docker-dev

erinyoung:
I think the steering committee of StaPH-B would be a good idea.

kapsakcj (approved):

PR #693, opened 2023-07-18
Create CONTRIBUTING.md

erinyoung:
Contributing readme files seem to be popular, so I made one. 

All this one does it send the user to https://staphb.org/docker-builds/contribute/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+Thank you for your interest in contributing to the StaPH-B/docker-builds repository! 
+
+Please see [https://staphb.org/docker-builds/contribute/](https://staphb.org/docker-builds/contribute/) for more information on how to contribute.


kapsakcj:
This new file is a bit "hidden" to most users. It may not be obvious where to find this file, and most importantly the link within.

Would you be open to the idea of removing this file & add another Heading at the bottom of the main `/README.md` (below the Authors/Maintainers Heading? We could list the link to https://staphb.org/docker-builds/contribute/ in this new section in the `/README.md`

erinyoung:
CONTRIBUTING.md is a specific file that gets picked up by certain tools, so it need to be in the base directory if we're going to use it. It may end up in the right sidebar with the description, license, code of conduct, etc in the future.

CONTRIBUTING.md is intended to 
1) signal to others that "outsider" contributions are accepted
2) gives instructions on how to contribute to the github repository

We already have a webpage and documents that cover contributing topics, so I'd rather link to those than copy and paste them into CONTRIBUTING.md 

Other repos that use it (https://github.com/search?q=CONTRIBUTING.md&type=repositories)
- https://github.com/leachim6/hello-world
- https://github.com/GoogleCloudPlatform/Template
- https://github.com/openwrt/packages

kapsakcj:
Ah, I did not know that. Thank you for the info and links. I agree with you, let's keep `/CONTRIBUTING.md` in that case

kapsakcj (approved):

PR #790, opened 2023-11-13
bedtools version 2.31.1

erinyoung:
There's a new version of bedtools, although this update seems to deal with installation issues (read more here https://github.com/arq5x/bedtools2/releases/tag/v2.31.1).

For this, I copied the Dockerfile from 2.31.0 and changed the version ARG. Everything else is the same.

Pull Request (PR) checklist:


bedtools/2.31.1/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+
+### keeping old installation cmds here in case we want to install via pre-compiled binary again in the future ###
+# RUN cd /usr/local/bin && \
+#     wget https://github.com/arq5x/bedtools2/releases/download/v${BEDTOOLS_VER}/bedtools.static && \
+#     mv bedtools.static bedtools && \
+#     chmod +x bedtools && \
+#     mkdir /data
+
+### test stage ###
+FROM builder as test
```
Kincekara, line 52: "FROM app as test"? 
We need to test the app to see any missing libraries in the app stage. So, the test stage needs to be after the app stage.

bedtools/2.31.1/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+
+# test scripts expect to be run from the bedtools root dir
+WORKDIR /bedtools2-${BEDTOOLS_VER}
+
+# commenting out ulimit command in test/test.sh (gitpod complains)
+# run tests included with bedtools code
+RUN sed -i 's/ulimit/#ulimit/g' test/test.sh && \
+make test
+
+### final app stage ###
```
Kincekara, line 65: This part needs to be done before the test stage to be used in the test stage.

Kincekara:
I believe those mistakes come from the previous image. Although it can work this way, fixing it in this PR would be better.

bedtools/2.31.1/Dockerfile
Kincekara: You will need to download the test files again in the test stage.

erinyoung: bedtools has a lot of subcommands, but the internal tests of bedtools has a dependency from builder.

I think when we did this one before, we tested it with some files after.

Let me get this one up to standard practice, though.

bedtools/2.31.1/Dockerfile
erinyoung: I've decided to move this into the builder stage and then I added a test to the test stage

erinyoung:
@Kincekara , Thank you for your suggestions!

I have made some adjustments by combined the old test stage into the builder stage - because I do like how extensively it gets tested.

I also moved the labels into the app stage and created a test stage built from app. 

Let me know if I should make any other changes!

bedtools/2.31.1/Dockerfile
Kincekara: That makes sense!

Kincekara:
@erinyoung Thank you for all your adjustments! Moving the internal test to the builder stage and creating a separate test is a great idea. Overall, the image looks fantastic.

Kincekara (approved):

Kincekara:
Thank you for your PR! 
You can check the deploy status from here: https://github.com/StaPH-B/docker-builds/actions/runs/6961689944

PR #796, opened 2023-11-14
GA for keeping Freyja updated

erinyoung:
Freyja's databases update frequently, but I think we can keep up with a github action (GA).

This GA
1. Identifies the most recent Dockerfile for Freyja to use for building and deployment
  A. This is also where the version is extracted
2. Builds that Dockerfile to the 'test' stage
3. Extracts the demix database version
4. Builds and deploys the image to the 'app' stage with the tag of `<version from most recent dockerfile>_<demix database version>` with the steps copied from the manual deploy workflow
  A. We could also add a date or something to the tag if we wanted to be really specific.

I haven't tested this completely (because of reasons), but if this works we could apply similar principles to other repositories that need to be updated periodically (i.e. MLST).


erinyoung:
@kapsakcj , I'm not in any particular hurry, so this can wait until the next meeting.

Also, I'm sure there are typos in it (GA are difficult to test), but I thought about it this morning and decided to create the PR before I lost the idea.

kapsakcj (approved):
Let's merge this and test it out. Thank you Erin!

PR #860, opened 2024-02-01
Update freyja to have static freyja version

erinyoung:
This PR removes the dynamic way I was originally looking for the most-recent version of Freyja

My original thought was to look for the most recent Freyja dockerfile, and extract the freyja version from that.

When I run the following locally
```
git log -1 --name-only --format=%cd --date=iso freyja/*/Dockerfile | grep Dockerfile | head -n 1
```
I get `freyja/1.4.8/Dockerfile`. I thought this would work.

When this is run in GA, the file is `freyja/1.2.1/Dockerfile`, which is the wrong file.

In lieu of this issue, I've commented out the old step and replaced it with one that  uses a static version. This means that every time we update Freya, we'll need to update this GA.

### DIFF
--- a/.github/workflows/update_freyja.yml
+++ b/.github/workflows/update_freyja.yml
@@ -35,17 +35,28 @@ jobs:
           
       - name: pull repo
         uses: actions/checkout@v3
-        
-      - name: get latest version of freyja in docker-builds repo
-        id: latest_version
-        run: | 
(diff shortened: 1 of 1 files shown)


Kincekara:
I am merging this. I hope it will work. 🤞

Kincekara (approved):

PR #884, opened 2024-02-21
Remove cache in freyja GA

erinyoung:
I've removed the lines for caching. I think this, in general, saves space in our GA repo, but I believe the cache is stopping Freyja from being created anew each day.

Right now the Freyja database tags are not updating, suggesting that the new databases aren't getting pulled.


Kincekara (approved):

Kincekara:
merging... 🤞

PR #902, opened 2024-03-11
MLST GA context instead of file

erinyoung:
https://github.com/StaPH-B/docker-builds/actions/runs/8236077721 failed because of a build error. The way it was written before made it so the test couldn't be copied into the image. 

I've replaced "file" with "context" which _should_ fix that issue.


Kincekara (approved):
🤞

PR #1083, opened 2024-10-28
update pango aliasor weekly

erinyoung:
This isn't a Dockerfile, it's a github action. I've had this on my todo list for a long time, but I need pango aliasor to update with the pangolin lineages. 

I was thinking weekly would be more than enough.

This github action 
1. builds the docker image to test
2. Then pushes the 'app' stage to quay and dockerhub with both the `latest` and `0.3.0-{date}` tags


kapsakcj (approved):
LGTM. I'll merge and if we need to make any further changes, you can adjust and submit another PR

PR #1085, opened 2024-10-29
fix bbtools deployment error

Kincekara:
This should fix the deployment error of bbtools 39.10.


erinyoung:
It looks like it finally made it to dockerhub! https://hub.docker.com/r/staphb/bbtools/tags

PR #1138, opened 2024-12-19
updating checkm2 docker build

taylorpaisie:
Pull Request (PR) checklist:


taylorpaisie:
I am creating this PR for the Checkm2 docker container.  I know the same version is already on master but I was building this container for our workflow already and thought I would submit and get some feedback for the container.  Let me know what you think!

Kincekara:
@taylorpaisie What is the reason for trying to update this image? Do you want to add its database to the image? If there is an issue in the current image, please submit an issue first. So we can avoid similar problems in future releases.

First of all, we don't overwrite the existing images except on very rare occasions. I can break people's pipelines and validations. 
Secondly, we try to avoid adding large databases to images. Because it significantly increases the loading time, decreases efficiency, and may cause problems in cloud applications. 

However, we can still add it if you need it as is. Please change your PR name to something like **checkm2/1.0.2-{current db name or version}** first. So, we can add it without overwriting the current image. 

Is it possible to use Jammy as the base image?

I'm just curious, have you compared the image sizes of micromaba and staged builds?

taylorpaisie:
The purpose of me creating this new image is to not use micromamba in the build.  I realize that micromamba creates a smaller image size, but I was also under the impression that micromamba, or anything conda related, in these docker builds should be avoided.

erinyoung:
I think it would be great to use a slimmer image. When we changed how things were installed for quast, we add a `-slim` to directory path and version tag (i.e. 5.3.0-slim).

I think we need to keep the original dockerfile unchanged, and then copy these edits into a new Dockerfile in a `build-files/checkm2/1.0.2-slim` subdirectory.

Or something like it. 

Maybe `focal` instead of `slim`?

Kincekara:
@erinyoung I do agree we should keep the original dockerfile unchanged. 
This image is not smaller than the current one. On the contrary, it is much larger since it includes ~3.1 GB of diamond database.
We can add it with a database name or date like bakta or busco images.

PR #1186, opened 2025-01-28
adding pangolin version 4.3.1 with pdata 1.32

erinyoung:
There's a new pdata version!!!

It appears that the pangolin version is the same. I tried updating the version of usher to 0.6.4, but encountered errors.

I changed a lot of tabs/spacing, so it looks like I changed the entire file. My apologies.

I moved where PATH is defined so that the "MAMBA_DOCKERFILE_ACTIVATE" lines could be removed.
I updated the pdata version arg to 1.32

Mostly what I did was add/fix the hyperlinks in the main README.md

I think in the future we can stop using 'sed' on the environment file, but I'm hoping for a "quick" turnaround time on this image.

Pull Request (PR) checklist:
- [X] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/build-files/samtools/1.15` )
- [X] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number in build-files (i.e. `docker-builds/build-files/spades/3.12.0/Dockerfile`)
   - [X] (optional) All test files are located in same directory as the Dockerfile (i.e. `build-files/shigatyper/2.0.1/test.sh`)
- [X] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `docker-builds/build-files/spades/3.12.0/README.md`)


kapsakcj:
>I think in the future we can stop using 'sed' on the environment file, but I'm hoping for a "quick" turnaround time on this image.

If we remove the sed lines then we are no longer pinning the versions of those dependencies. The versions chosen would be up to the conda env solver. So I'd recommend we keep the sed commands to ensure versions are pinned

I removed the sed commands on line 52-56, built the docker image, and it installed usher 0.5.4 which is quite old and this caused errors when running pangolin in the test stage. It installed scorpio 0.3.19 and constellations 0.1.12 which is good, but I still would rather pin these versions.

Usher 0.6.4 isn't yet available on bioconda unfortunately https://github.com/bioconda/bioconda-recipes/pull/52300

kapsakcj:
I was a little worried that removing the `MAMBA_DOCKERFILE_ACTIVATE` would cause issues but that seems to not be the case as I tested this docker image in a Terra/WDL workflow and everything ran smoothly. 

Since this variable is removed, the container is now defaults to the `base` conda environment instead of the `pangolin` conda environment. But you've circumvented that by hardcoding the path to the conda env binaries into the $PATH.

kapsakcj (approved):

kapsakcj:
thank you for the PR! deploy workflow here: https://github.com/StaPH-B/docker-builds/actions/runs/13039574668

PR #1368, opened 2025-07-22
Update Dockerfile_mamba

erinyoung:
The dockerfile that we recommend people use as a template is a little old, so I've tweaked some things.

- updated the base image to the current one of today
- added micrombamba list as a suggestion in the test stage
- removed some environmental variables (these are normally not needed when PATH is set correctly)


Kincekara (approved):

Kincekara:
Thank you for updating this template. I love the changes! I also added `-f` to micromamba clean line

erinyoung:
I forgot that one! Thank you!

PR #1405, opened 2025-08-21
Fixing and Adding hyperlinks

erinyoung:
There is no dockerfile in this PR.

Instead, this updates the main readme.md
- fixes hyperlinks
- adds hyperlinks

This should create merge conflicts with everything else. (But... since there aren't a lot of PRs... this seems like a better time than most...)


Kincekara:
That's great! Thank you for fixing all the links.

Issue #1615, opened 2026-04-07
[Repo Request]: update github actions


kapsakcj:
### What functionality is this repo missing?

I noticed some warnings in the github actions logs:

<img width="2757" height="895" alt="Image" src="https://github.com/user-attachments/assets/2d32c599-b989-465b-b01e-de181b1c7818" />

We should consider upgrading to the latest versions, like `actions/cache@v5` : https://github.com/marketplace/actions/cache

`actions/checkout@v6` https://github.com/marketplace/actions/checkout

and others

Looks like this should be completed before June 2nd or we may face errors/issues with the automated GHActions workflows

kapsakcj:
more info here https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Issue #1249, opened 2025-03-05
Removing conda defaults channel


Kincekara:
I suggest removing the conda defaults channel from all future images wherever possible to avoid changing the license terms of Anaconda. I think conda-forge can be replaced with defaults channel even if the developers suggest defaults in their installation documents.

kapsakcj:
totally agree on this, let's discourage the use of `-c defaults` for conda-based installs

Issue #750, opened 2023-09-06
[Repo Request]: New Templates


erinyoung:
### What functionality is this repo missing?

We have several templates available in [dockerfile-template](https://github.com/StaPH-B/docker-builds/tree/master/dockerfile-template), but there could be more that may be useful. 

Additional templates:
[ ] R-packages (like ggtree)
[ ] tools that are snakemake wrappers (like ncov-recombinant)

kapsakcj:
Might be nice to have a `rust` template too? Not super important but would be nice for rust-based binfie tools!

Could start with the Dockerfile added for sylph in PR #816

kapsakcj:
Perhaps uses this as a starting point for snakemake-based tools? https://github.com/snakemake/snakemake/blob/main/Dockerfile

kapsakcj:
Could be useful for creating a dockerfile for a snakemake workflow: https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#containerization-of-conda-based-workflows

Issue #1600, opened 2026-03-18
dockerhub repo Overview pages are out of date


kapsakcj:
Not sure how often people look at the "Overview" tab on dockerhub, but they are out-of-date and kind of a mess.

Way back when we started this project, dockerhub would inherit the `/README.md` from this github repo and it would update every time we used dockerhub's build infrastructure. But doesn't stay up-to-date unless you build an image with their infrastructure. And since we don't do that anymore, they are terribly out of date and misleading to users.

Example: https://hub.docker.com/r/staphb/any2fasta Overview page is from 4 years ago.

Some of the Overview pages were manually updated by myself or other maintainers.

Anyways, they're all over the place and I'm thinking we should update them with a big link to the GitHub repo as it's documentation will be the most up-to-date going forward.

I'm not sure if there's a way to programmatically update the Overview pages but we should look for a solution like that since there are 245(!) Overview pages that need updating.

PR #1616, opened 2026-04-07
Updates for Dockerfile templates

erinyoung:
This PR contains 
1. New dockerfile templates. 
  a. The templates include a CMD line, new labels, creation of a non-root user, and other updates.
  b. There's a LOT more of them for images built in rust, R, and java
2. Updated github actions for these new files.
  a. Build to test is now split into three sections to assist in code review : builder (optional), app, and test
  b. Adjusted for new labels
  c. Explicit testing for nextflow deps
  d. Added linting and security testing (set up to be ignored right now)
  e. Created a bot to assist in comparing two files


erinyoung:
Closes https://github.com/StaPH-B/docker-builds/issues/1615
Closes https://github.com/StaPH-B/docker-builds/issues/750
Closes https://github.com/StaPH-B/docker-builds/issues/1249
Attempt at https://github.com/StaPH-B/docker-builds/issues/1600

dockerfile-template/Dockerfile
```diff
@@ -8,85 +10,96 @@
...
+# NOTE: The "org.opencontainers" labels are REQUIRED to pass StaPH-B CI tests.
+
+# Required Metadata Labels
+# These follow the OCI Image Spec. They help users and automated systems 
+# identify what is inside the container.
+LABEL \
+      # The date this image was last updated (Manually update this!)
+      org.opencontainers.image.created="2024-05-20" \
+      # The exact base image used for the 'app' stage (Must match the FROM instruction above)
+      org.opencontainers.image.base.name="ubuntu:jammy" \
```
Kincekara: ubuntu:noble

erinyoung:
This is what these look like in action: https://github.com/erinyoung/docker-builds/pull/8

dockerfile-template/Dockerfile
erinyoung: Looks like most of the org.opencontainer.image.base.name were incorrect. I think they are all fixed now.

Kincekara:
> This is what these look like in action: [erinyoung#8](https://github.com/erinyoung/docker-builds/pull/8)

oh, that's awesome!

.github/workflows/build_and_test_pr_dockerfile.yml
```diff
@@ -0,0 +1,203 @@
...
+          
+      - name: Validate Nextflow Dependencies
+        continue-on-error: true
+        id: nextflow
+        run: |
+          # Define IMAGE locally so it is available in this shell session
+          IMAGE="${{ inputs.tool_name }}:app"
+
+          FAILED=0
+          if docker run --rm $IMAGE ps --version > /dev/null 2>&1; then
```
kapsakcj, line 132: for this line and similar lines below, would it be helpful to remove `/dev/null 2>&1`? Just thinking that it might be useful to see that output in the github actions log in case we need to see the error message (or conversely, see the version of `ps`, `sed`, etc.

dockerfile-template/Dockerfile
```diff
@@ -8,85 +10,96 @@
...
+      # The version of THIS Dockerfile (increment if you change the build logic)
+      org.opencontainers.image.revision="1" \
+      # A human-readable description of the tool
+      org.opencontainers.image.description="A brief description of what this tool does" \
+      # The project homepage or GitHub repository
+      org.opencontainers.image.url="https://github.com/StaPH-B/docker-builds" \
+      # Where users can find instructions or manuals for the software
+      org.opencontainers.image.documentation="https://github.com/StaPH-B/docker-builds" \
+      # Link to the source code used to build the image (which is often a GitHub url)
+      org.opencontainers.image.source="https://github.com/StaPH-B/docker-builds" \
```
kapsakcj, line 45: reminder to double-check that this label `org.opencontainers.image.source` as well as the label `org.opencontainers.image.url` are accurate

If one of them will always link back to this github repo, we should state that in the comment so users know to keep it static

dockerfile-template/Dockerfile
```diff
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
```
kapsakcj, line 1: Just curious - what does this comment line do? Is there a meaning behind this?

.github/workflows/build_and_test_pr_dockerfile.yml
erinyoung: The error has to be routed to stdout instead of stderr, or the step fails at `docker run --rm $IMAGE ps --version` and won't continue on.

dockerfile-template/Dockerfile
erinyoung: We can remove this if we want to. Our Dockerfiles are pretty basic.

This tells Docker which parser/feature set to use when reading the Dockerfile, specifically it instructs Docker to use the BuildKit Dockerfile frontend (docker/dockerfile). The `:1` means version 1 of that syntax. 


It must appear at the very top of the Dockerfile before any other instructions.

I am not clear about all the features of using this line, but it enables modern Dockerfile features that aren’t available in the legacy parser, such as `- RUN --mount=type=cache`.


There's supposed to be improved performance with BuildKit.

erinyoung:
This PR was too large and had too many steps and ideas.

I have split it into several PR:
https://github.com/StaPH-B/docker-builds/pull/1616
https://github.com/StaPH-B/docker-builds/pull/1652
https://github.com/StaPH-B/docker-builds/pull/1653

PR #1651, opened 2026-04-21
New templates

erinyoung:
This PR contains some example templates that we can point to when people are creating new Dockerfiles.

- Dockerfile : basic 'app' and 'test' stage
- Dockerfile_builder : adds a 'builder' stage and corresponding 'COPY' layer to app
- Dockerfile_java : adds a java builder stage
- Dockerfile_mamba : updates the base image
- Dockerfile_micromamba_builder : adds a micromamba 'builder' stage, but is otherwise similar to Dockerfile_builder
- Dockerfile_python3 : is basically 'Dockerfile' with the python3-slim base
- Dockerfile_R : admittedly I haven't tested this, but it has a builder stage and some handy R stuff for installation
- Dockerfile_rust : is basically Dockerfile_builder using a 'rust' base image for 'builder'
- README.md : has added text for the micromamba and pip dependencies (which are optional)

### DIFF
--- a/dockerfile-template/Dockerfile
+++ b/dockerfile-template/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
+
 ##### ----- ----- ----- ----- ----- ----- ----- ----- ----- ----- ----- #####
 ##### Thank you for using this Dockerfile template!                     #####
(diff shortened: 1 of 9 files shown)


erinyoung:
Closes https://github.com/StaPH-B/docker-builds/issues/750

PR #1653, opened 2026-04-21
New Find Dockerfile in PR and Build

erinyoung:
These github action yml files are intended to replace `test-PR-dockerfiles.yml`.

- find_new_pr_dockerfile.yml uses `tj-actions/changed-files` to identify if a PR contains a new Dockerfile. It then creates a json of all the Dockerfiles in the PR and will pass them to
  - build_and_test_pr_dockerfile.yml : builds a dockerfile to 'builder' (optionally), 'app', and 'test' stages. Will then check the 'app' stage for labels, the workdir, a non-root user, check for some common nextflow depdencencies, run trivy, and generate a report back to the PR for each dockerfile.
  - lint_pr_dockerfile.yml : does not build anything, instead looks at the file for linting with hadolint, and some common layers (e.g. CMD)
 
 Most of the text on test-PR-dockerfiles.yml is for testing purposes, and can probably be removed if we're confident they work.

### DIFF
--- a/.github/workflows/build_and_test_pr_dockerfile.yml
+++ b/.github/workflows/build_and_test_pr_dockerfile.yml
@@ -1,0 +1,201 @@
+name: Build and Test
+
+on:
+  workflow_call:
+    inputs:
+      context_path:
+        required: true
+        type: string
+      tool_name:
+        required: true
+        type: string
+      version:
+        required: true
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@main
+      - uses: docker/setup-buildx-action@master
+
+      - name: Check for Builder Stage
+        id: check-builder
+        run: |
+          if grep -qi "AS builder" "${{ inputs.context_path }}/Dockerfile"; then
+            echo "exists=true" >> $GITHUB_OUTPUT
+          else
+            echo "exists=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Build Builder Stage
+        if: steps.check-builder.outputs.exists == 'true'
+        uses: docker/build-push-action@master
+        with:
+          context: ${{ inputs.context_path }}
+          target: builder
+          load: true
+          tags: ${{ inputs.tool_name }}:builder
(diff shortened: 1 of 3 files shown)


erinyoung:
Closes https://github.com/StaPH-B/docker-builds/issues/1615

erinyoung:
Closes https://github.com/StaPH-B/docker-builds/issues/1249

kapsakcj:
Plan: 
- merge the PR and then test the new deploy workflow on an image that needs to be deployed
  - If new deploy workflow fails, then we can still use the old “manual-deploy” workflow

.github/workflows/build_and_test_pr_dockerfile.yml
```diff
@@ -0,0 +1,226 @@
...
+          WORKDIR=$(docker inspect --format '{{.Config.WorkingDir}}' "$IMAGE")
+          
+          if [ "$WORKDIR" != "/data" ]; then
+            echo "Error: WORKDIR is set to '$WORKDIR'. It must be set exactly to /data."
+            exit 1
+          else
+            echo "Success: WORKDIR is correctly set to /data"
+          fi
+
+      - name: Validate Non-Root User
```
kapsakcj, line 106: If I understand this part correctly, this would require all dockerfiles to declare a non-root user with `USER <username>` dockerfile directive

I know that it's best practice in terms of security to declare non-root user, but I fear that this could cause issues with file permissions & perhaps other things such as re-usability of the image (like using an image as a base for another image, e.g. VADR).

Docker images where files inside the image (e.g. a database of FASTA files or something) are root owned would need to have their permissions adjusted so all users can read (and maybe write?) to those files. It would add additional complexity to docker images.

I don't know for sure, but definitely something to consider and we should discuss

erinyoung: Yes! Best practice is to add a non-root user, but I don't think we'll be locking `root` from the user. For the example of VADR, we would need to set the USER to root at the beginning of the dockerfile, and then switch to a non-root user at the end again. We do something similar to this when we use micromamba as a base.

PR #1683, opened 2026-05-13
Falco 1.3.0

RaheelSyedAhmed:
Pull Request (PR) checklist:
- [X] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/build-files/samtools/1.15` )
- [X] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number in build-files (i.e. `docker-builds/build-files/spades/3.12.0/Dockerfile`)
   - [ ] (optional) All test files are located in same directory as the Dockerfile (i.e. `build-files/shigatyper/2.0.1/test.sh`)
- [X] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `docker-builds/build-files/spades/3.12.0/README.md`)

Update to Falco directory to include version 1.3.0 of Falco, added a URL for documentation on all flags for Falco in the program README.md, and updated the main README.md to include version 1.3.0 in the versions dropdown entry for Falco.


RaheelSyedAhmed:
Hello Dr. @Kincekara, I made a quick change to the apt-get installation in the app phase. The prior edition installs libhts-dev in both the builder and the app phase. I changed this to only install libraries shown through `ldd /usr/local/bin`. This saves roughly 500 MB when the Docker image is built.

Would you like me to keep the libhts-dev installation for clarity and continuity, or keep the essential libraries in the app stage to limit size? See [d021558](https://github.com/StaPH-B/docker-builds/commit/d021558684086b8af9527347476c65eb1ca7a9b1) lines 44-51 for further detail.

Kincekara:
The compressed size of the previous image is about 42 MB, so it’s unlikely that adding `libhts-dev` alone accounts for an extra 500 MB. The difference is more likely coming from the apt layer, since `--no-install-recommends` isn’t available during the builder stage.

You can verify the uncompressed image size with the `docker images` command in your workspace. As a rule of thumb, the compressed size is usually a bit less than half of the uncompressed size.

There is a trade-off between Dockerfile complexity and image size. While smaller images are generally preferable for security and efficiency, a simpler Dockerfile can be the better long‑term choice if the size difference is not significant, since it makes future updates easier to maintain.

Kincekara:
You can keep your changes if the image works properly.

RaheelSyedAhmed:
Thank you for the input Dr. Incekara! I'll wait for the checks to complete and swap this from a draft PR to a regular PR.

Kincekara:
@staphb-dockerbuilds-diff falco 1.2.5 1.3.0

Kincekara (approved):

PR #1843, opened 2026-09-18
adding seqsender version 1.5.1

erinyoung:
There's a new version of seqsender!

According to the [release notes](https://github.com/CDCgov/seqsender/releases/tag/v1.5.1) this fixes an outdated python package.

At this moment in time, I'm not 100% sure if this will actually run. The environment file that is included in the repo does not use the required beartype and many other dependencies. I have adjusted it in this PR, and it looks like it _should_ work fine. It builds in any case, but only because I used GEMINI as errors arose.

This is the summary from GEMINI:

## Key Changes

* **Upgraded Python Runtime to 3.10 & Sanitized `env.yaml`**
  * **Change:** Added a `sed` step to strip hardcoded `py38` build hashes from `env.yaml` and forced `python=3.10` during `micromamba install`.
  * **Reason:** `src/settings.py` uses PEP 585 generic type annotations (`list[str]`), which throws `TypeError: 'type' object is not subscriptable` on Python 3.8. Stripping build hashes allows `micromamba` to build modern Python 3.10 binary packages.

* **Removed Legacy PyPI `typing` Package**
  * **Change:** Removed `pip install typing` and deleted `typing` from package installation lines.
  * **Reason:** The `typing` package on PyPI is an obsolete backport for Python 3.4 that overwrites standard library files, resulting in `ImportError: cannot import name 'Annotated' from 'typing'`.

* **Pinned Dependency Constraints (`pandera=0.18.3`, `numpy<2`, `beartype`)**
  * **Change:** Explicitly passed `pandera=0.18.3`, `numpy<2`, and `beartype` to `micromamba install`.
  * **Reason:**
    * **`pandera=0.18.3`**: Prevents resolution to `pandera` v0.19+, which removed the `from pandera import pandera` internal module imported by `src/upload_log.py`.
    * **`numpy<2`**: `pandera 0.18.3` relies on `np.string_`, which was deprecated and removed in NumPy 2.0 (`AttributeError`).
    * **`beartype`**: Directly required by `src/__init__.py` but absent from `env.yaml`.

* **Corrected Test Stage CLI Flags & Source Paths**
  * **Change:** Switched the test generation flag from `-g` to `--genbank` and set `WORKDIR` to `/seqsender-${SEQSENDER_VERSION}` during test execution.
  * **Reason:** `seqsender.py` rejected `-g` as an unrecognized argument. Executing from the source tree allows `seqsender.py` to locate its bundled `test_data/` template directories.

Pull Request (PR) checklist:
- [x] Include a description of what is in this pull request in this message.
- [x] The dockerfile successfully builds to a test target for the user creating the PR. (i.e. `docker build --tag samtools:1.15test --target test docker-builds/build-files/samtools/1.15` )
- [x] Directory structure as name of the tool in lower case with special characters removed with a subdirectory of the version number in build-files (i.e. `docker-builds/build-files/spades/3.12.0/Dockerfile`)
   - [x] (optional) All test files are located in same directory as the Dockerfile (i.e. `build-files/shigatyper/2.0.1/test.sh`)
- [x] Create a simple container-specific [README.md](https://github.com/StaPH-B/docker-builds/blob/master/.github/workflow-templates/readme-template.md) in the same directory as the Dockerfile (i.e. `docker-builds/build-files/spades/3.12.0/README.md`)
   - [x] If this README is longer than 30 lines, there is an explanation as to why more detail was needed
- [x] Dockerfile includes the recommended [LABELS](https://github.com/StaPH-B/docker-builds/blob/master/dockerfile-template/Dockerfile#L8-L18) 
- [x] Main [README.md](https://github.com/StaPH-B/docker-builds/blob/master/README.md) has been updated to include the tool and/or version of the dockerfile(s) in this PR
- [x] [Program_Licenses.md](https://github.com/StaPH-B/docker-builds/blob/master/Program_Licenses.md) contains the tool(s) used in this PR and has been updated for any missing


erinyoung:
@staphb-dockerbuilds-diff seqsender 1.3.93 1.5.1

